Issue #3, opened 2019-07-10
Paired Result - Multiple Regions


abhijit945:
When all but one data set is toggled off from the legend (note: a data set can contain paired values), the region(s) should appear face up

**Actual**

![image](https://user-images.githubusercontent.com/4908886/61313689-9234bd00-a7c0-11e9-96b0-ee1b4db87ca3.png)

**Expected**

It should be consistent to Line Graph.

![image](https://user-images.githubusercontent.com/4908886/61313670-89dc8200-a7c0-11e9-8646-648719ded1df.png)

jodybutts:
@abhijit945 do you mean that in a Paired Results graph with more than 2 variables we should not allow any regions to display? 

In the particular graph you referenced, we are not showing more than one at the same time, only via hover. Plus, when I turn off all but one graph, the region is not displaying for the one that remains.

I think we could treat it just the same as Regions > Line > Multiple line and use the default grey for each, but also display the region when there is only a single variable toggled on. 

What are your thoughts?

abhijit945:
I am not clear what you meant in the original issue on Cerner GH (I moved this issue from there).

jodybutts:
So this was something I had previously logged? I don't remember the context behind it, sorry. Is it possible to see the original chain still?

jodybutts:
After further discussion via Teams, we discussed that only one change is necessary: when all but one data set is toggled off from the legend (note: a data set can contain paired values), the region(s) should appear face up. This is to ensure consistency with the normal line multi-line scenario.

abhijit945:
Updated the defect description.

PR #49, opened 2019-07-18
Update: same region show up

Prashant-Kamboj:

[] Documentation update
[] Bug fix
[] New functionality
[X] Changes an existing functionality
[] Other, please explain:

Changed the logic to show the region, if region is same for multi-graph it will show up by default.
**Add any screenshots**
image

image


src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -188,10 +191,101 @@ class PairedResult extends GraphContent {
...
         if (utils.notEmpty(this.dataTarget.regions)) {
-            // If graph has more than 1 content then we dont load the regions
+            const regionList = this.dataTarget.regions;
+            const values = this.dataTarget.values;
+
+            //check if all region are there with respect to value
+            if (graph.config.isPairedDataProper === true) {
+                for (let i = 0; i < values.length; i++) {
+                    // eslint-disable-next-line max-depth
+                    for (const key in values[i]) {
```
Dinesh94Singh: Do we need to loop through keys? I think low, mid and high are the only possible keys.

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -188,10 +191,101 @@ class PairedResult extends GraphContent {
...
+            }
+            if (
+                localRegionSameFlag === true &&
+                graph.content.length <= 1 &&
+                graph.config.isPairedDataProper === true
+            ) {
+                globalregionData.pop();
+                globalregionData.push(regionItm[0]);
+            }
+            //Globale rgion comparision
```
Dinesh94Singh: nit: Global region

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -188,10 +191,101 @@ class PairedResult extends GraphContent {
...
+                    for (let i = 0; i < region.length; i++) {
+                        // eslint-disable-next-line max-depth
+                        if (regionItm.length === 0) {
+                            regionItm.push(region[i]);
+                        }
+                        // eslint-disable-next-line max-depth
+                        if (regionItm.length > 0) {
+                            // eslint-disable-next-line max-depth
+                            if (
+                                regionItm[0].start === region[i].start &&
```
Dinesh94Singh: Might probably use https://www.npmjs.com/package/deep-equal package for comparison of objects. Think there should be one in the helpers as well.

src/main/js/helpers/region.js
```diff
@@ -17,6 +17,70 @@ import utils from "./utils";
...
+ */
+const compareRegionDataLine = (region) => {
+    if (globalregionData.length > 0) {
+        if (
+            globalregionData[0].start === region.start &&
+            globalregionData[0].end === region.end &&
+            globalregionData[0].axis === region.axis
+        ) {
+            return true;
+        } else {
```
Dinesh94Singh: unnecessary use of else block, return directly

src/main/js/helpers/region.js
```diff
@@ -17,6 +17,70 @@ import utils from "./utils";
...
+        globalregionData.length > 0 &&
+        graphConfig.isPairedDataProper === true
+    ) {
+        if (
+            globalregionData[0].start === region.start &&
+            globalregionData[0].end === region.end &&
+            globalregionData[0].axis === region.axis
+        ) {
+            return true;
+        } else {
```
Dinesh94Singh: Unnecessary use of else block in - line 63 and 66, return directly

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: we have to iterate and check if the region for each (high, mid and low ) is there or not bcz values can be multiple, and i don't want to hard code keys to check.

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: i will correct it

src/main/js/helpers/region.js
Prashant-Kamboj: isRegionSame me be set true for one load and for the second load the region is not same for line then we have to return false.

src/main/js/helpers/region.js
Prashant-Kamboj: isRegionSame me be set true for one load and for the second load the region is not same for pairedResult then we have to return false.

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: ok i will use it for the object comparison

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -188,10 +191,101 @@ class PairedResult extends GraphContent {
      * @inheritDoc
      */
     resize(graph) {
+        const regionItm = [];
```
abhijit945: Sorry but I dont think this logic would be necessary. Atleast a bulk of it. 
Consider this: 
The region element already has all the data you need (Except axis), for instance in a `multiple line single region scenario` ([here](https://engineering.cerner.com/carbon-graphs/#/regions/line/multiple-line)):

The 2 regions are:

image

When you get the data for these elements:

image

The reason why the axis is not available is because the default is Y Axis.

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -188,10 +191,101 @@ class PairedResult extends GraphContent {
...
+        let localRegionSameFlag = false;
+
         if (utils.notEmpty(this.dataTarget.regions)) {
-            // If graph has more than 1 content then we dont load the regions
+            const regionList = this.dataTarget.regions;
+            const values = this.dataTarget.values;
+
+            //check if all region are there with respect to value
+            if (graph.config.isPairedDataProper === true) {
```
abhijit945: In case of a Paired Result its pretty much the same case but an extra step is necessary, currently we are only putting the data of `start`, `end` and `axis` to the regions but along with that you will also put the `key` (most likely [here](https://github.com/cerner/carbon-graphs/blob/master/src/main/js/controls/PairedResult/helpers/helpers.js#L605)). What this will allow is that you will get the context of which region is belonging to what content. 

As always, the logic would be to show the region until one of them is different.

| Content 1 | Content 2 | Content 3 | Result   |
| --------- | --------- | --------- | -------- |
| R1        | R2        | R3        | Hide All |
| R1        | R1        | R2        | Hide All |
| R1        | R1        | R1        | Show     |
| R1        | -         | -         | Hide All |
| R1        | R1        | -         | Hide All |
| -         | -         | -         | N/A      |

src/main/js/helpers/region.js
Dinesh94Singh: Right, if it doesn't go into the if block, then simply return ```false``` would be sufficient. If it goes into if block, it returns ```true```

```
        if (
            globalregionData[0].start === region.start &&
            globalregionData[0].end === region.end &&
            globalregionData[0].axis === region.axis
        ) {
            return true;
        } else {
           return false;
       }   
```
is the same as 

```
        if (
            globalregionData[0].start === region.start &&
            globalregionData[0].end === region.end &&
            globalregionData[0].axis === region.axis
        ) {
            return true;
        } 
        return false;
```

src/main/js/controls/PairedResult/PairedResult.js
Dinesh94Singh: Looking elsewhere how carbon handles such scenarios, it looks like they use the keys. @abhijit945 - what do you think?

abhijit945: Essentially you dont need to do this.
Consider this, as a graph the only thing we need to consider is what regions are drawn. There is no difference between how a line region and a paired graph region are rendered, they are just `rect` with different `y` pos. 

For instance: 

* For 2 regions belonging to separate contents (_Line 1_ has `R1` and _Line 2_ has `R2`)
    * Compare `R1` element's `datum()` and `R2` element's `datum()`
* For 3 regions belonging to separate contents (_Line 1_ has `R1` and _Paired Result 1_ has `R2` and `R3`)
    * After R2 and R3 is rendered, we just sequentially compare the first element `R1` with all the rest (`R2 and R3` in our case), since there is only one scenario where we show the regions `i.e. R1, R2 and R3 should all have same regions in same axis`.

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: This logic is only for paired Result and as we discussed  in the previous PR there were limitation for picking the elements from the dom.

src/main/js/helpers/region.js
Prashant-Kamboj: ok i will do the change

dev/app.js
```diff
@@ -68,7 +69,8 @@ import {
     renderLineY2Axis,
     renderLineYHidden,
     renderMultiLine,
-    renderMultiLineRegion
+    renderMultiLineRegion,
+    renderMultiLineSameRegion
```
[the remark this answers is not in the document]
Prashant-Kamboj: ok i will do the change

dev/app.js
```diff
@@ -20,7 +20,8 @@ import { renderColorsExample } from "./examples/controls/colors";
 import {
     renderCombinationBar,
     renderCombinationRegion,
-    renderCombinationSimple
+    renderCombinationSimple,
+    renderCombinationSameRegion
```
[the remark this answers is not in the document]
Prashant-Kamboj: ok i will do the change

dev/app.js
```diff
@@ -532,6 +534,16 @@ renderSiteApp(
                         }
                     ]
                 },
+                {
+                    pathname: "/regions/multi-line",
+                    children: [
+                        {
+                            pathname: "/regions/multi-line/identical",
+                            content: renderMultiLineSameRegion,
+                            title: "Same Region ShowUp"
```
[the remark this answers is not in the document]
Prashant-Kamboj: ok i will do the change

dev/app.js
```diff
@@ -575,6 +587,11 @@ renderSiteApp(
                             pathname: "/regions/combination/simple",
                             content: renderCombinationRegion,
                             title: "Simple"
+                        },
+                        {
+                            pathname: "/regions/combination/identical",
+                            content: renderCombinationSameRegion,
+                            title: "Same Region ShowUp"
```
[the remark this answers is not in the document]
Prashant-Kamboj: ok i will do the change

dev/examples/controls/combination.js
```diff
@@ -68,3 +68,44 @@ export const renderCombinationBar = (id) => {
...
+        mid: [
+            {
+                axis: "y",
+                start: 6,
+                end: 100
+            }
+        ]
+    };
+    const combGraph = Carbon.api.graph(
+        getDemoData(`#${id}`, "LINE_PAIR_COMB_DEFAULT")
```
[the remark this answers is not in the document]
Prashant-Kamboj: OK i will make the common word for the axis label too.

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: We can not go with the dom approach  since in the case if region is not provided it will not create the region for it, that is why we switched to this approach.

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: we discussed  in the previous PR all the scenarios if high, mid and low are provided in the value,
But in the region we done have the region for say low in the case it will be missing in the dom too and if me make the decision on the dom bases that was giving us an improper decision (as the in the previous logic in previous PR) since the region for low is missing even though all other regions are same.

If all the region are same one region is missing the picking from dom will make the region show up which is not what we need same for the line too if we don't even provide the region for say line and for paired we have given all region same then in that case too DOM will give us improper information.

abhijit945: I am not sure I understand this correctly. Are you concerned with some contents not having regions?

src/main/js/controls/PairedResult/PairedResult.js
abhijit945: I am not for this for 2 reasons: 
* If we deep equal on an input object then in future you cannot include an object property without breaking this functionality, if it doesnt apply to condition here. For instance, if we decide a `label` or `onClick` property. 
* Pulling in additional dependencies

src/main/js/controls/PairedResult/PairedResult.js
abhijit945: That works even today. We just hide all the regions, at that point. 
https://github.com/cerner/carbon-graphs/blob/master/src/main/js/controls/Line/Line.js#L167

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: yes i am concerned with some contents not having regions.

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: so you want me not to use deep-equal

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: its for Paired not for line.

src/main/js/controls/PairedResult/PairedResult.js
Dinesh94Singh: Don't we have a comparator sought of thing in our helpers @abhijit945, which we can use?

src/main/js/controls/PairedResult/PairedResult.js
abhijit945: This is already handled isnt it? The 3rd content in [this](https://engineering.cerner.com/carbon-graphs/#/regions/line/multiple-line) example doesnt have a region so everything is hidden.

Prashant-Kamboj: No this is not handled. 
It's like if the content is > 1  call hideallregion which is not proper .

Prashant-Kamboj: for that example even if the region is same for 2 content also it will not show the region.  
No case was handled earlier.

src/main/js/helpers/region.js
```diff
@@ -208,14 +208,51 @@ const shouldHideAllRegions = (regionList, graphTargets) =>
  * @returns {boolean} true if displayed targets is equal to 1, false otherwise
  */
 const isSingleTargetDisplayed = (graphTargets) => graphTargets.length === 1;
+/**
+ * Check all the regions within a graph is same or not
+ * @private
+ * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
+ * @returns {Object} d3 svg path
```
abhijit945: You are not returning anything in this function.

src/main/js/helpers/region.js
```diff
@@ -208,14 +208,51 @@ const shouldHideAllRegions = (regionList, graphTargets) =>
  * @returns {boolean} true if displayed targets is equal to 1, false otherwise
  */
 const isSingleTargetDisplayed = (graphTargets) => graphTargets.length === 1;
+/**
+ * Check all the regions within a graph is same or not
```
abhijit945: You are doing more than checking all the regions in this function, you are hiding them as well. Instead you could do just the checking here and just call hideAll if you decide to hide. 
They are shown by default anyway, so you dont need to do, aria-hidden -> false for all the regions, if I am not wrong.

src/main/js/helpers/region.js
```diff
@@ -246,14 +283,26 @@ const toggleRegion = (canvasSVG, key) =>
...
  * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
  * @returns {Object} d3 svg path
  */
-const processRegions = (shownTargets, canvasSVG) =>
-    isSingleTargetDisplayed(shownTargets)
-        ? toggleRegion(canvasSVG, ...shownTargets)
+const processRegions = (graphConfig, canvasSVG) =>
+    isSingleTargetDisplayed(graphConfig.shownTargets)
+        ? toggleRegion(canvasSVG, ...graphConfig.shownTargets)
+        : checkRegion(graphConfig, canvasSVG);
```
abhijit945: `checkAllRegions`.
Previously this line read like -> If single region then toggle this region or hide all regions. 
Now it will read like -> If single region then toggle or check all regions.

src/main/js/helpers/region.js
```diff
@@ -208,14 +208,51 @@ const shouldHideAllRegions = (regionList, graphTargets) =>
  * @returns {boolean} true if displayed targets is equal to 1, false otherwise
  */
 const isSingleTargetDisplayed = (graphTargets) => graphTargets.length === 1;
+/**
+ * Check all the regions within a graph is same or not
+ * @private
+ * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
+ * @returns {Object} d3 svg path
+ */
+const checkRegionSame = (canvasSVG) => {
```
abhijit945: You can call this `areRegionsSame()`

src/main/js/helpers/region.js
```diff
@@ -246,14 +283,26 @@ const toggleRegion = (canvasSVG, key) =>
...
+/**
+ * Checks region for legend iteam click
+ * @private
+ * @param { Object } graphConfig - graph propertiy needed to check the isRegionSame property
+ * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
+ * @returns { callback } - hiderAllegions or showAllRegions
+ */
+const checkRegion = (graphConfig, canvasSVG) => {
+    graphConfig.shownTargets.length > 1 && graphConfig.isHideAllRegion === false
+        ? checkRegionSame(canvasSVG)
```
abhijit945:
```
if (graphConfig.shownTargets.length > 1 && !graphConfig.isHideAllRegion && areRegionsSame())
	return;
}
hideAllRegions(canvasSVG);
```

src/main/js/helpers/region.js
```diff
@@ -208,14 +208,51 @@ const shouldHideAllRegions = (regionList, graphTargets) =>
...
+    }
+};
 /**
  * Hides all the regions within a graph
  * @private
  * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
  * @returns {Object} d3 svg path
  */
-const hideAllRegions = (canvasSVG) =>
+const hideAllRegions = (canvasSVG) => {
```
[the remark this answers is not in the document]
Prashant-Kamboj: braces  are not needed i will remove it

src/main/js/helpers/region.js
```diff
@@ -208,14 +208,51 @@ const shouldHideAllRegions = (regionList, graphTargets) =>
...
+ * @private
+ * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
+ * @returns {Object} d3 svg path
+ */
+const checkRegionSame = (canvasSVG) => {
+    const regionData = [];
+    const regions = canvasSVG.selectAll(`.${styles.region}`);
+    regions[0] !== undefined
+        ? regions[0].forEach((element) => {
+              regionData.push(element.__data__);
```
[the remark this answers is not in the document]
Prashant-Kamboj: yes right i will correct it

src/main/js/helpers/region.js
Prashant-Kamboj: no they are not shown by default i tried to do so but it did not worked.

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -188,18 +189,49 @@ class PairedResult extends GraphContent {
      * @inheritDoc
      */
     resize(graph) {
+        let isPairedDataProper = true;
         if (utils.notEmpty(this.dataTarget.regions)) {
-            // If graph has more than 1 content then we dont load the regions
```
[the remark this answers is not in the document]
Prashant-Kamboj: Because now the logic is different i will change the comment to 
**// If graph has more than 1 content then we compare the regions if same or not**

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -188,18 +189,49 @@ class PairedResult extends GraphContent {
...
+            const regionList = this.dataTarget.regions;
+            const values = this.dataTarget.values;
+
+            //check if all region are there with respect to value
             if (graph.content.length > 1) {
-                hideAllRegions(graph.svg);
+                for (let i = 0; i < values.length; i++) {
+                    // eslint-disable-next-line max-depth
+                    for (const key in values[i]) {
```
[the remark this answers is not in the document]
Prashant-Kamboj: but need it  because values is an array for each value we have to check if the region is proper or not( for each value high , mid or low ) region is also there which we can not check for this constant(constants.PAIR_ITEM_TYPES).

src/main/js/helpers/region.js
abhijit945: "they are not shown by default" Wait what? Where is it hidden by default?

src/main/js/controls/PairedResult/PairedResult.js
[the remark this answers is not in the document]
abhijit945:
```
// If graph has more than 1 content, we compare the regions if they are identical and hide if even atleast one of them is not. 
```

src/main/js/controls/PairedResult/PairedResult.js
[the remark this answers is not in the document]
abhijit945: What are you trying to achieve with this? 

I pulled down your code and tried it myself: 

My `values[0]` is
image

You are checking if high, mid and low are keys are present in the region, which obviously it wont.

image


Furthermore, I removed the nested for loop entirely and just have this
```javascript
 //check if all region are there with respect to value
            if (graph.content.length > 1) {
                if (
                    graph.config.isHideAllRegion === false
                ) {
                    checkRegionSame(graph.svg);
                    graph.config.isHideAllRegion = false;
                } else {
                    hideAllRegions(graph.svg);
                    graph.config.isHideAllRegion = true;
                }
            }
```
Yet, all your unit tests in PairedResult pass:
image

src/main/js/helpers/region.js
```diff
@@ -246,14 +283,26 @@ const toggleRegion = (canvasSVG, key) =>
...
  */
-const processRegions = (shownTargets, canvasSVG) =>
-    isSingleTargetDisplayed(shownTargets)
-        ? toggleRegion(canvasSVG, ...shownTargets)
+const processRegions = (graphConfig, canvasSVG) =>
+    isSingleTargetDisplayed(graphConfig.shownTargets)
+        ? toggleRegion(canvasSVG, ...graphConfig.shownTargets)
+        : checkRegion(graphConfig, canvasSVG);
+/**
+ * Checks region for legend iteam click
```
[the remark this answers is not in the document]
Prashant-Kamboj: i will correct it

src/main/js/helpers/region.js
```diff
@@ -246,14 +283,26 @@ const toggleRegion = (canvasSVG, key) =>
...
+ * @private
+ * @param { Object } graphConfig - graph propertiy needed to check the isRegionSame property
```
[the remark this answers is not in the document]
Prashant-Kamboj: i will correct it

src/main/js/helpers/region.js
```diff
@@ -246,14 +283,26 @@ const toggleRegion = (canvasSVG, key) =>
...
+ * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
+ * @returns { callback } - hiderAllegions or showAllRegions
```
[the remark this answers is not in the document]
Prashant-Kamboj: i will correct it

src/main/js/controls/Graph/GraphConfig.js
```diff
@@ -123,6 +123,7 @@ export const processInput = (input, config, type) => {
     }
     config.shownTargets = [];
     config.hasCriticality = false;
+    config.isHideAllRegion = false;
```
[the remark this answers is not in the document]
Prashant-Kamboj: i will change it to **shouldHideAllRegion**

src/main/js/helpers/region.js
Prashant-Kamboj: ok will change it to **@returns {undefined} - returns nothing**

src/main/js/helpers/region.js
Prashant-Kamboj: yes i will rename it to **areRegionsSame()**

src/main/js/helpers/region.js
Prashant-Kamboj: it's working it's showing by default you were right

src/main/js/helpers/region.js
Prashant-Kamboj: ok i will rename it to **checkAllRegions**

src/main/js/helpers/region.js
```diff
@@ -208,14 +208,51 @@ const shouldHideAllRegions = (regionList, graphTargets) =>
...
+    regions[0] !== undefined
+        ? regions[0].forEach((element) => {
+              regionData.push(element.__data__);
+          })
+        : regionData.push();
+
+    if (regionData.length === 1 || regionData.length === 0) {
+        canvasSVG.selectAll(`.${styles.region}`).attr("aria-hidden", true);
+    } else {
+        for (let i = 0; i < regionData.length - 1; i++) {
```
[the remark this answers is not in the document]
Prashant-Kamboj: ok i will use forEach loop.

src/main/js/helpers/region.js
```diff
@@ -208,14 +208,51 @@ const shouldHideAllRegions = (regionList, graphTargets) =>
...
+const checkRegionSame = (canvasSVG) => {
+    const regionData = [];
+    const regions = canvasSVG.selectAll(`.${styles.region}`);
+    regions[0] !== undefined
+        ? regions[0].forEach((element) => {
+              regionData.push(element.__data__);
+          })
+        : regionData.push();
+
+    if (regionData.length === 1 || regionData.length === 0) {
```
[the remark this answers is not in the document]
Prashant-Kamboj: no it's not needed i will remove this line.

src/main/js/controls/Line/Line.js
```diff
@@ -11,8 +11,8 @@ import {
     createRegion,
     hideAllRegions,
     removeRegion,
-    shouldHideAllRegions,
-    translateRegion
+    translateRegion,
```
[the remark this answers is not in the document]
Prashant-Kamboj: yes i will remove the function

src/main/js/controls/Line/Line.js
```diff
@@ -158,13 +158,18 @@ class Line extends GraphContent {
...
+        if (utils.notEmpty(this.dataTarget.regions)) {
+            if (
+                graph.content.length > 1 &&
+                graph.config.isHideAllRegion === false
+            ) {
+                checkRegionSame(graph.svg);
+                graph.config.isHideAllRegion = false;
+            }
+        }
+        if (utils.isEmpty(this.dataTarget.regions)) {
```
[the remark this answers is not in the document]
Prashant-Kamboj: yes i will move it to else part

src/main/js/controls/PairedResult/PairedResult.js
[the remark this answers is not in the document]
Prashant-Kamboj: it's is because all the condition for region are proper i have not given those scenarios  where region is missing for any value, that is why all test case are passing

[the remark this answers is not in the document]
Prashant-Kamboj: what i am trying to achieve is i am checking for each value if region for each for each key in the value is there or not.

example - vlaue[0] - is an object having value for high, mid and low and i am checking that region is also there for all high mid and low, if for any value it is missing we will set the parameter  ispairedData proper as false and call hideAllRegions.

src/main/js/helpers/region.js
Prashant-Kamboj: but we set , aria-hidden -> false for all the regions because when we click on legend we are calling the areRegionSame() so when all legends are clicked (config.showTarget.length becomes 0) and all regions are hidden so to show again we are call the function areRegionSame() after hiding we have to show again that is why we are setting the aria-hidden -> false.

src/main/js/controls/PairedResult/PairedResult.js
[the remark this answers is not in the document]
Prashant-Kamboj: i will change the comment

src/main/js/helpers/region.js
```diff
@@ -208,14 +208,51 @@ const shouldHideAllRegions = (regionList, graphTargets) =>
...
+        canvasSVG.selectAll(`.${styles.region}`).attr("aria-hidden", true);
```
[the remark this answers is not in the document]
Prashant-Kamboj: yes right i have remove the line.

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -188,18 +189,49 @@ class PairedResult extends GraphContent {
...
+                        // eslint-disable-next-line max-depth
+                        if (!regionList.hasOwnProperty(key)) {
```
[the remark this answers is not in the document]
Prashant-Kamboj: Values is an array we can use array.every() on it but values[i] is an object we can not use array function on it.

src/main/js/controls/PairedResult/helpers/helpers.js
```diff
@@ -431,7 +432,9 @@ const processRegions = (graphContext, config, canvasSVG, { key }) => {
             config.shownTargets.indexOf(key) > -1
         );
     } else {
-        hideAllRegions(canvasSVG);
+        config.isHideAllRegion === false && config.shownTargets.length > 0
```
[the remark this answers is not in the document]
Prashant-Kamboj: the if condition is checking only for graphcontent.length === 1 so for combination(Line,PairedResult) it will always go in the else and adding the config.shownTargets.length > 0 is because if all legends are clicked ( config.shownTargets.length = 0)and single legend is clicked again (config.shownTargets.length = 1) then also we have to show the region if it is same ( only combination)

src/main/js/helpers/region.js
Prashant-Kamboj: ok i will make the change

src/test/unit/controls/Line/Line-spec.js
```diff
@@ -2203,6 +2203,60 @@ describe("Line", () => {
                 });
             });
         });
+        describe("Show region face-up when there are multiple regions with multiline with same data", () => {
```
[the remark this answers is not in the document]
Prashant-Kamboj: ok i will make it short and discrete for describe.

src/main/js/controls/PairedResult/PairedResult.js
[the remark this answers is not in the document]
abhijit945: So in short you are hiding the regions for high and low if there is not for high, low and mid? Is that a new requirement?

src/main/js/helpers/region.js
abhijit945: Yeah I am not understanding this at all. Is there a unit test scenario for this?

src/main/js/helpers/region.js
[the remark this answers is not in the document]
abhijit945: I would use `some` you wont need to use `break`.

src/main/js/helpers/region.js
Prashant-Kamboj: now i am handling that case in the processRegion only if areRegionSame is true then show region only for legend click. 

Yes for line we have toggleRegion if (showTarget === 1) but for pairedResult we don't have toggle for the condition (showTarget === 1) in that case we are checking (graphContext.content === 1).

src/main/js/helpers/region.js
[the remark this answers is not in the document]
Prashant-Kamboj: i am using every in the case for i am checking if region is same always
 `const isSame = regionData.every((element) => {        
            if( compare.start === element.start &&
                compare.end === element.end &&
                compare.axis === element.axis) {
                    return true;                  
            }      
            return false;
        })
        return isSame;`

src/main/js/controls/PairedResult/PairedResult.js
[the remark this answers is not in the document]
Prashant-Kamboj: it's very simple values=[
{ high:{},mid:{},low:{}}
]

regions={
high:{},
low:{}
}
so for low since there is no region we should hideAllRegions.

src/test/unit/controls/Line/Line-spec.js
```diff
@@ -2203,6 +2203,60 @@ describe("Line", () => {
...
+                    "false"
+                );
+                expect(regionsElement[0].getAttribute("aria-describedby")).toBe(
+                    `region_${inputPrimary.key}`
+                );
+                expect(regionsElement[1].getAttribute("aria-describedby")).toBe(
+                    `region_${inputSecondary.key}`
+                );
+            });
+        });
```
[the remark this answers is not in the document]
Prashant-Kamboj: yes i will add more test case from the truth table.

src/main/js/helpers/region.js
```diff
@@ -246,14 +283,26 @@ const toggleRegion = (canvasSVG, key) =>
...
+        ? toggleRegion(canvasSVG, ...graphConfig.shownTargets)
+        : checkRegion(graphConfig, canvasSVG);
+/**
+ * Checks region for legend iteam click
+ * @private
+ * @param { Object } graphConfig - graph propertiy needed to check the isRegionSame property
+ * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
+ * @returns { callback } - hiderAllegions or showAllRegions
+ */
+const checkRegion = (graphConfig, canvasSVG) => {
```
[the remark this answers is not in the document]
Prashant-Kamboj: yes i will make the change.

src/main/js/helpers/region.js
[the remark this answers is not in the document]
abhijit945: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

Let me know which is optimal for your use-case.

src/main/js/controls/PairedResult/helpers/helpers.js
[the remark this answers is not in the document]
abhijit945: I was more expecting a unit test for this. But this is also connected to the bug
https://github.com/cerner/carbon-graphs/issues/3
Ill re-review this once you post your changes.

src/main/js/helpers/region.js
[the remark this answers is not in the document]
Prashant-Kamboj: every is optimal for your use-case.

src/test/unit/controls/PairedResult/PairedResult-spec.js
```diff
@@ -3112,6 +3114,124 @@ describe("PairedResult", () => {
...
+                    "true"
+                );
+                expect(regionsElement[4].getAttribute("aria-hidden")).toBe(
+                    "true"
+                );
+                expect(regionsElement[5].getAttribute("aria-hidden")).toBe(
+                    "true"
+                );
+                expect(regionsElement[6].getAttribute("aria-hidden")).toBe(
+                    "true"
```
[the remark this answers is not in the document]
Prashant-Kamboj: i did typeof regionsElements it returned object so we can not use every since it's an object not an array.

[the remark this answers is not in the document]
abhijit945: image

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array

src/main/js/helpers/region.js
```diff
@@ -197,23 +197,43 @@ const translateRegion = (scale, config, regionGroupSVG) =>
...
- * @param {Array} graphTargets - List of all the items in the Graph
- * @returns {boolean} true if displayed targets is equal to 1, false otherwise
+ * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
+ * @returns {boolean} - returns true is regions are same else false
  */
-const isSingleTargetDisplayed = (graphTargets) => graphTargets.length === 1;
+const areRegionSame = (canvasSVG) => {
+    const regionData = [];
+    const regions = canvasSVG.selectAll(`.${styles.region}`).data();
+    regions !== undefined
```
abhijit945: I dont think this does anything. You can just update the whole function to
```js
const regions = canvasSVG.selectAll(`.${styles.region}`).data();
const compare = regions[0];
return !regions.some(
    (element) =>
        compare.start !== element.start ||
        compare.end !== element.end ||
        compare.axis !== element.axis
);
```

src/main/js/helpers/region.js
```diff
@@ -197,23 +197,43 @@ const translateRegion = (scale, config, regionGroupSVG) =>
...
+        ? regions.forEach((element) => {
+              regionData.push(element);
+          })
+        : regionData.push();
+    const compare = regionData[0];
+    const isSame = regionData.every((element) => {
```
abhijit945: The reason I asked you to use `some` instead of `every` is that you wont get immediate feedback. 
If you use `every` then even if you have a non-matching item, the function would just continue to put every item through that check. 
`Some` will fail fast.

src/test/unit/controls/Line/Line-spec.js
```diff
@@ -2215,6 +2216,101 @@ describe("Line", () => {
                 });
             });
         });
+        describe("Check if region same for multiline with same dataset", () => {
```
abhijit945: Dataset is a term used for data values, in your case the dataset is not same is it? `inputPrimary` and `inputSecondary` have different values. Might want to update the description here.

src/test/unit/controls/Line/Line-spec.js
```diff
@@ -2215,6 +2216,101 @@ describe("Line", () => {
...
+                    `region_${inputPrimary.key}`
+                );
+                expect(regionsElement[1].getAttribute("aria-describedby")).toBe(
+                    `region_${inputSecondary.key}`
+                );
+                expect(regionsElement[2].getAttribute("aria-describedby")).toBe(
+                    `region_${inputThird.key}`
+                );
+            });
+            it("If any region is missing hide regions", () => {
```
abhijit945: Instead of saying what you expect here. You can have it as
`Shows region if one or more are identical`
`Hides region if one or more are not identical`
`Hides region if one or more is missing`

src/test/unit/controls/Line/helpers.js
```diff
@@ -108,6 +108,13 @@ export const inputSecondary = {
     },
     values: valuesDefault
 };
+export const inputThird = {
```
abhijit945: `inputTertiary`

src/main/js/controls/PairedResult/helpers/helpers.js
```diff
@@ -449,7 +450,15 @@ const processRegions = (graphContext, config, canvasSVG, { key }) => {
             config.shownTargets.indexOf(key) > -1
         );
     } else {
-        hideAllRegions(canvasSVG);
+        if (
+            config.shouldHideAllRegion === false &&
+            config.shownTargets.length > 0 &&
+            areRegionSame(canvasSVG)
+        ) {
+            canvasSVG.selectAll(`.${styles.region}`).attr("aria-hidden", false);
```
abhijit945: Could we use `checkAllRegions` function here?

src/test/unit/controls/Line/Line-spec.js
Prashant-Kamboj: ok i will update the description.

src/test/unit/controls/Line/Line-spec.js
Prashant-Kamboj: ok i will update the it description as above.

src/main/js/helpers/region.js
```diff
@@ -197,23 +197,43 @@ const translateRegion = (scale, config, regionGroupSVG) =>
...
+ * Check all the regions within a graph is same or not
  *
  * @private
- * @param {Array} graphTargets - List of all the items in the Graph
- * @returns {boolean} true if displayed targets is equal to 1, false otherwise
+ * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
+ * @returns {boolean} - returns true is regions are same else false
  */
-const isSingleTargetDisplayed = (graphTargets) => graphTargets.length === 1;
+const areRegionSame = (canvasSVG) => {
```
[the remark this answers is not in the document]
Prashant-Kamboj: areRegionsIdentical is also good

src/main/js/helpers/region.js
```diff
@@ -256,14 +276,34 @@ const toggleRegion = (canvasSVG, key) =>
  * * If only 1 target is displayed -> show the region using unique data set key
  *
  * @private
- * @param {Array} shownTargets - Targets/data sets that are currently displayed in graph
+ * @param {object} { shownTargets, shouldHideAllRegion } - Targets/data sets that are currently displayed in graph
```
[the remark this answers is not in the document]
Prashant-Kamboj: ok i will update the JSDoc for destructured objects

src/test/unit/controls/Line/helpers.js
Prashant-Kamboj: ok i will rename

src/main/js/helpers/region.js
```diff
@@ -197,23 +197,43 @@ const translateRegion = (scale, config, regionGroupSVG) =>
             return getRegionHeight(d3.select(this), d, scale, config);
         });
 /**
- * Decides if regions needs to be hidden based on count of displayed targets in graph
+ * Checks if only 1 content item is present in the graph
  *
  * @private
- * @param {Array} regionList - List of regions to be shown within graph
+ *
```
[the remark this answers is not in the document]
Prashant-Kamboj: i will remove the empty line

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -191,17 +192,43 @@ class PairedResult extends GraphContent {
...
-                hideAllRegions(graph.svg);
+                const isRegionThere = (value) => {
+                    for (const key in value) {
+                        if (!regionList.hasOwnProperty(key)) {
+                            return false;
+                        }
+                    }
+                    return true;
+                };
+                const isPairedDataProper = values.every(isRegionThere);
```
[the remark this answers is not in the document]
Prashant-Kamboj: I have already added one test case for this for multi-paired with missing region.

src/main/js/controls/PairedResult/helpers/helpers.js
```diff
@@ -449,7 +450,15 @@ const processRegions = (graphContext, config, canvasSVG, { key }) => {
             config.shownTargets.indexOf(key) > -1
         );
     } else {
-        hideAllRegions(canvasSVG);
+        if (
```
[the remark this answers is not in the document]
Prashant-Kamboj: I will update the JSDoc with this "Regions would be checked if they are identical before hiding them."

src/main/js/helpers/region.js
Prashant-Kamboj: yes right i will replace it with above.

src/main/js/helpers/region.js
Prashant-Kamboj: yes right

src/main/js/controls/PairedResult/helpers/helpers.js
Prashant-Kamboj: I don't think because  we are also checking for config.shownTargets.length > 0 but in checkAllRegion we are checking for config.shownTargets.length > 1 so we can't.

src/main/js/controls/PairedResult/helpers/helpers.js
```diff
@@ -449,7 +450,15 @@ const processRegions = (graphContext, config, canvasSVG, { key }) => {
...
+            config.shouldHideAllRegion === false &&
+            config.shownTargets.length > 0 &&
+            areRegionSame(canvasSVG)
+        ) {
+            canvasSVG.selectAll(`.${styles.region}`).attr("aria-hidden", false);
+        } else {
+            hideAllRegions(canvasSVG);
```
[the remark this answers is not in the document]
Prashant-Kamboj: yes right.

src/main/js/controls/PairedResult/PairedResult.js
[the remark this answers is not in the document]
abhijit945: Which unit test is that?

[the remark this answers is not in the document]
Prashant-Kamboj: ->  it("Hide region if region is missing for any value(high, mid, low)") 

line no - 3204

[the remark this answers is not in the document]
abhijit945: Doesnt all unit test pass even with the condition and every line associated to `isPairedDataProper` removed? 

```js
if (graph.content.length > 1) {
    if (
        graph.config.shouldHideAllRegion === false &&
        areRegionSame(graph.svg)
    ) {
        graph.config.shouldHideAllRegion = false;
    } else {
        hideAllRegions(graph.svg);
        graph.config.shouldHideAllRegion = true;
    }
}
```

image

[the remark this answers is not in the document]
Prashant-Kamboj: it is because the condition i have given for regionMissing in that region are not same so areRegionsSame is returning false so that is why the test cases are  passing if you change the parameters  of regionMissing to same as the multiRegionSameData then this will fail in that case.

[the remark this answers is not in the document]
abhijit945: So you are going to fix the test? Sorry, what is the resolution here?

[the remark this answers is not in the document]
Prashant-Kamboj: it is working properly, no fix is need for this.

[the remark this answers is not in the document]
Prashant-Kamboj: I made some changes to the test case

abhijit945:
TravisCI build is failing at
image
Can you look into it.

Prashant-Kamboj:
yes i am working on that only.

src/main/js/controls/PairedResult/PairedResult.js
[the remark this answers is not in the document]
abhijit945: This can be further optimised to 
```js
const regionList = this.dataTarget.regions;
const values = this.dataTarget.values;
const isRegionMappedToAllValues = (value) =>
    Object.keys(value).every((v) =>
        regionList.hasOwnProperty(v)
    );
const isPairedDataProper = values.every(isRegionMappedToAllValues);
```

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -191,17 +192,43 @@ class PairedResult extends GraphContent {
...
+
+                if (
+                    isPairedDataProper === true &&
```
abhijit945:
```js
if (
    isPairedDataProper &&
    !graph.config.shouldHideAllRegion &&
    areRegionsIdentical(graph.svg)
)
```

src/main/js/controls/PairedResult/helpers/helpers.js
```diff
@@ -448,6 +449,12 @@ const processRegions = (graphContext, config, canvasSVG, { key }) => {
             `region_${key}`,
             config.shownTargets.indexOf(key) > -1
         );
+    } else if (
+        config.shouldHideAllRegion === false &&
```
abhijit945:
```js
!config.shouldHideAllRegion &&
config.shownTargets.length > 0 &&
areRegionsIdentical(canvasSVG)
```

src/main/js/controls/PairedResult/PairedResult.js
```diff
@@ -191,17 +192,43 @@ class PairedResult extends GraphContent {
      */
     resize(graph) {
         if (utils.notEmpty(this.dataTarget.regions)) {
-            // If graph has more than 1 content then we dont load the regions
+            // If graph has more than 1 content, we compare the regions if they are identical and hide if even atleast one of them is not.
+            const regionList = this.dataTarget.regions;
```
abhijit945: Move this into line 199 so that it can only be created when there are multiple content.

src/main/js/helpers/region.js
abhijit945: This is still pending.

src/main/js/controls/PairedResult/PairedResult.js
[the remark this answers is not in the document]
Prashant-Kamboj: ok i will optimise the code further

src/main/js/helpers/region.js
Prashant-Kamboj: i made the change

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: ok i will move the line to 199.

src/main/js/controls/PairedResult/helpers/helpers.js
Prashant-Kamboj: i will correct it

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: yes i will make the change

src/main/js/controls/PairedResult/helpers/helpers.js
```diff
@@ -449,7 +450,15 @@ const processRegions = (graphContext, config, canvasSVG, { key }) => {
             config.shownTargets.indexOf(key) > -1
         );
     } else {
-        hideAllRegions(canvasSVG);
+        if (
+            config.shouldHideAllRegion === false &&
+            config.shownTargets.length > 0 &&
+            areRegionSame(canvasSVG)
```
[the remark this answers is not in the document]
Prashant-Kamboj: added test for this covering all the three line of code.

dev/examples/controls/bar.js
```diff
@@ -202,7 +202,7 @@ export const renderBarTimeSeries = (id) => {
     );
     return barTime;
 };
-export const renderBarTimeSeriesWithDateline = (id) => () => {
+export const renderBarTimeSeriesWithDateline = (id) => {
```
abhijit945: 💯

src/main/js/controls/PairedResult/PairedResult.js
abhijit945: You can shorten this to `isPairedDataProper` instead of `isPairedDataProper === true`.

src/main/js/controls/PairedResult/helpers/helpers.js
```diff
@@ -448,10 +449,25 @@ const processRegions = (graphContext, config, canvasSVG, { key }) => {
...
+        areRegionsIdentical(canvasSVG)
+    ) {
+        canvasSVG.selectAll(`.${styles.region}`).attr("aria-hidden", false);
     } else {
         hideAllRegions(canvasSVG);
     }
 };
+/**
+ * Checks if region is missing for any key in the value (high, mid or low)
+ *
```
abhijit945: Lets add the **why** along with **what** here. 
You can add something like: `We are doing this so that the regions are hidden if one of regions for Paired Result data pairs are not provided, i.e. if regions for "high" and "low" are provided and the values contain data for "high", "mid" and "low", all regions would be hidden`.

Prashant-Kamboj: i added the comment to describe more.

src/main/js/controls/PairedResult/PairedResult.js
Prashant-Kamboj: made the change.

abhijit945:
PR will be merged when the subtask in JIRA for testing is closed.

Prashant-Kamboj:
Documentation is done waiting for the update in the requirement to add the requirement number.

abhijit945:
Merged. Good job 🎉 🎉 
The version number will be updated here post release

PR #55, opened 2019-07-24
Update: Load and unload batch of tracks (fixes #51)

### LINKED ISSUE #51: Load and unload a batch of rows
Carbon currently supports the addition and removal of rows through loadContent() and unloadContent() functions, but these only accept a single row at a time. It could potentially increase performance if these functions were able to load and unload a batch of rows. 

We tested carbon to load rows in batch and experienced  significantly faster load time on large loads/unloads. For our use case we tested with Gantt graphs.

habelk:

[] Documentation update
[] Bug fix
[] New functionality
[X] Other, please explain: Improve Gantt performance  to load/unload tracks 

Utilized https://github.com/cerner/carbon-graphs/blob/master/src/main/js/helpers/loadUtil.js#L18 to load and unload tracks in batches.

See  https://github.com/cerner/carbon-graphs/issues/51

**Add any screenshots**
The screen shots captures loading 100 tracks and then attempting to unload all tracks after a one second delay. You can see it is around 8 to 10 times faster.

**Before**
Load_unload_existing

**After**
Load_unload_batch


src/main/js/controls/Gantt/Gantt.js
```diff
@@ -287,20 +290,22 @@ class Gantt extends Construct {
...
+     * @param {object|Array} content - Gantt content to be removed
      * @returns {Gantt} - Gantt instance
      */
     unloadContent(content) {
-        const index = this.tracks.indexOf(content.key);
-        if (index < 0) {
-            throw new Error(errors.THROW_MSG_INVALID_OBJECT_PROVIDED);
-        }
-        this.trackConfig[index].unload(this);
+        contentLoadHandler(content, (i) => {
```
abhijit945: I am wondering if it will be safer for us to create a `contentUnloadHandler` in this case.

src/main/js/controls/Gantt/Gantt.js
```diff
@@ -287,20 +290,22 @@ class Gantt extends Construct {
...
+            const index = this.tracks.indexOf(i.key);
+            if (index < 0) {
+                throw new Error(errors.THROW_MSG_INVALID_OBJECT_PROVIDED);
+            }
+            this.trackConfig[index].unload(this);
+            this.tracks.splice(index, 1);
+            this.trackConfig.splice(index, 1);
+        });
```
abhijit945: Is there a same performance gain during unload as well? I am trying to figure what is the `n` value here and if we need to do this for other types of graph as well.

src/test/unit/controls/Gantt/Gantt-spec.js
```diff
@@ -2284,6 +2304,27 @@ describe("Gantt", () => {
...
+                expect(gantt.trackConfig.length).toBe(0);
+                expect(gantt.config.height).toBe(0);
+                expect(gantt.config.axis.y.domain.lowerLimit).toBe(0);
+                expect(gantt.config.axis.y.domain.upperLimit).toBe(1);
+                expect(gantt.config.canvasHeight).toBe(
+                    BASE_CANVAS_HEIGHT_PADDING
+                );
+                done();
+            }, TRANSITION_DELAY);
+        });
```
abhijit945: Can you update the documentation as well? We would want to advertise that the data can be loaded as array now. Refer [pie.md](https://github.com/cerner/carbon-graphs/blob/master/docs/controls/Pie.md#data) `Usage`, `JSON Properties` and `Structure`

src/main/js/controls/Gantt/Gantt.js
habelk: Agree. Do you prefer another file (unloadUtil) for this or just add to loadUtil.js

src/main/js/controls/Gantt/Gantt.js
habelk: Yes there is similar gains for unload as well. I think it is worth logging an issue to look at all other graphs.

src/main/js/controls/Gantt/Gantt.js
abhijit945: Ah good point. How about having the function `contentUnloadHandler` in the same file but rename the file to `constructUtils.js` (loadContent and unloadContent are Construct functions)

src/test/unit/controls/Gantt/Gantt-spec.js
habelk: done

PR #58, opened 2019-07-26
Fix: Axes shapes werent centered in Firefox (fixes #53)

### LINKED ISSUE #53: Y-Axis label icons not supported on Firefox browser for Multi-Line graph
**Description of the bug**

The Y-Axis label icons gets chopped off on Firefox browsers for multi-line graph. They will show up when you shrink it to smaller screen sizes but in wrong place and not all the labels will show up if there are <5 label icons in the graph. This works fine with other browsers(chrome, ie).

**Screenshots**

image
image

**Expected behavior**
image
image

abhijit945:

[] Documentation update
[X] Bug fix
[] New functionality
[] Other, please explain:

In Firefox, `Element.getBoundingClientRect` -> `height` was processed differently compared to Chrome or IE. We have moved away from the calculation for `getBoundingClientRect` at this point since the width would always increase when new contents are loaded. 
Now, we are simply re-positioning when a content is loaded or unloaded. [3]

**Add any screenshots**

_Firefox:_ 

_Chrome:_ 

[3]
When new content is loaded

image


abhijit945:
@Ashish1109 Can you take a look at this? Thanks.

src/main/js/helpers/axis.js
```diff
@@ -429,12 +429,12 @@ const getYAxisLabelShapeXPosition = (config) =>
...
  * @param {object} config - config object derived from input JSON
- * @param {Selection} labelShapeContainer - d3 element for label shape container
+ * @param {number} shapeCount - Number of shapes within shape container
  * @returns {number} Position for the label
  */
-const getYAxisLabelShapeYPosition = (config, labelShapeContainer) =>
+const getYAxisLabelShapeYPosition = (config, shapeCount) =>
     getYAxisLabelYPosition(config) +
-    labelShapeContainer.node().getBoundingClientRect().height / 2;
+    (shapeCount * constants.BASE_LABEL_ICON_HEIGHT_PADDING) / 1.5;
```
swatikode: Do we need to add a comment saying why we are dividing by 1.5?

PR #59, opened 2019-07-28
Fix: Regions not rendering combination with Paired Result (fixes #40)

### LINKED ISSUE #40: Fix: Region is not getting created for line if paired result is loaded first and then line.
**Describe the bug**
Region is not getting created for line if paired result is loaded first and then line, even if region is there in the line.

**To Reproduce**

Load paired result first on the Graph API canvas ( with region)
Load the line on the same Graph API canvas (with region)

**Expected behavior**
The region for the line should show in the carbon region if we load the line after paired result.

**Screenshots**
image

-   OS: [10.13.6 ]
-   Browser [chrome]

abhijit945:

[] Documentation update
[X] Bug fix
[] New functionality
[] Other, please explain:

Changed the way regions are injected from various contents that are not aware of other contents loading within the Graph API construct. 
Before we were loading all the values within an array as regions. 
i.e. 
**n** items in an array => **n** regions for a content
All regions are **rect** but PairedResult regions are loaded within a **g** element. This prevents the older logic from loading regions when used in a combination graph scenario.

**Add any screenshots**

image

image


abhijit945:
@Prashant-Kamboj Can you take a look at this? Thanks

PR #62, opened 2019-08-07
Update: Display Data Series In Legend when No Data Exists

NarasimhaShenoi:

[] Documentation update
[] Bug fix
[X] New functionality
[] Other, please explain:

If the consumer specifies the line as **panning mode** and no data is present for a data series for a given time frame, then the data series symbol and label will display within the legend but will display as disabled.  i.e. The label and symbol will appear in a dithered state and will not have the ability to be toggled on/off. 

*Add any screenshots**
Screen Shot 2019-08-07 at 2 56 46 PM (2)
Screen Shot 2019-08-07 at 2 56 38 PM (2)


src/main/js/core/BaseConfig/helper.js
```diff
@@ -29,7 +29,7 @@ export const validateBaseInput = (input) => {
     if (utils.isEmpty(input.key)) {
         throw new Error(errors.THROW_MSG_UNIQUE_KEY_NOT_PROVIDED);
     }
-    if (utils.isEmpty(input.values)) {
+    if (utils.isUndefined(input.values) || input.values === null) {
```
abhijit945: Do you want to do `!utils. isArray()` instead here?

src/main/js/helpers/legend.js
```diff
@@ -70,15 +70,17 @@ const loadLegendItem = (legendSVG, t, shownTargets, eventHandlers) => {
     validateLegendLabel(t.label);
     const text = getText(t.label.display);
     const index = shownTargets.indexOf(t.key);
+    const disabled =
+        !!t.label.isDisabled || (!!t.values && t.values.length == 0);
```
abhijit945: You can use `utils.isEmpty()` here? You already have validated it to be an array during Validation Phase anyway.

src/main/js/helpers/legend.js
```diff
@@ -70,15 +70,17 @@ const loadLegendItem = (legendSVG, t, shownTargets, eventHandlers) => {
     validateLegendLabel(t.label);
     const text = getText(t.label.display);
     const index = shownTargets.indexOf(t.key);
+    const disabled =
```
abhijit945: We are force disabling the legend item here, so might consider renaming this to `shouldForceDisableLegendItem`.

src/test/unit/controls/Bar/Bar-spec.js
```diff
@@ -70,7 +70,7 @@ describe("Bar", () => {
         });
         it("throws error when no values are provided", () => {
             const input = utils.deepClone(getInput(valuesDefault));
-            input.values = [];
+            input.values = undefined;
```
abhijit945: Can you add one more test to the specs to `Does not throw error when empty values array is provided`?

src/test/unit/controls/Line/Line-spec.js
```diff
@@ -70,9 +70,42 @@ describe("Line", () => {
         });
         it("throws error when no values are provided", () => {
             expect(() => {
-                graphDefault.loadContent(new Line(getInput([], false, false)));
+                graphDefault.loadContent(
+                    new Line(getInput(undefined, false, false))
+                );
             }).toThrowError(errors.THROW_MSG_NO_DATA_POINTS);
         });
+        it("display the legend when empty array is provided as input", () => {
```
abhijit945: Can you add this to the legend section instead? It will be easy to move if we ever decide to split the unit tests into separate files. (The files are getting long and hard to maintain)

Somewhere around [here](https://github.com/cerner/carbon-graphs/blob/8e61d42a04879ed922537f04c3d68f447bc725c5/src/test/unit/controls/Line/Line-spec.js#L943)

src/test/unit/controls/Line/Line-spec.js
```diff
@@ -70,9 +70,42 @@ describe("Line", () => {
...
+            const legendItems = legendContainer.children;
+            expect(legendContainer).not.toBeNull();
+            expect(legendContainer.tagName).toBe("UL");
+            expect(legendItems.length).toBe(1);
+            const legendItem = document.body.querySelector(
+                `.${styles.legendItem}`
+            );
+            expect(legendItem.getAttribute("aria-disabled")).toBe("true");
+        });
+        it("display the legend when values are provided", () => {
```
abhijit945: Can you add these legend tests to Paired Result and Bar as well?

abhijit945:
Code-wise it looks good, just need to add unit tests you created in Line to Paired Result and Bar specs as well.

NarasimhaShenoi:
I have added the test cases to the Paired Result and Bar specs.

src/main/js/helpers/legend.js
```diff
@@ -70,15 +70,17 @@ const loadLegendItem = (legendSVG, t, shownTargets, eventHandlers) => {
...
     const text = getText(t.label.display);
     const index = shownTargets.indexOf(t.key);
+    const shouldForceDisableLegendItem =
+        !!t.label.isDisabled || utils.isEmptyArray(t.values);
     const itemPath = legendSVG
         .append("li")
         .classed(styles.legendItem, true)
-        .attr("aria-selected", index > -1)
-        .attr("aria-disabled", !!t.label.isDisabled)
+        .attr("aria-selected", shouldForceDisableLegendItem || index > -1)
```
abhijit945: So when there is no data, this will be force-set to `true`? Is that what you want to do here? 
What happens in a scenario where, you have:
* Data is present from Range 0-40
* No Data is present from Range 41-50
* Data is present from Range 51-100

and you deselect your data set at 0-40? Just curious

NarasimhaShenoi: This code is confirming the style of the disabled element is the same across the application. 

Previously, When there is no data, by default (only in paired-result) 'aria-selected' going to be false. So, even after setting the disabled style, the icon shade used to be like unselected, without the shade in it. like below:
Screen Shot 2019-08-12 at 6 54 40 PM
Where as the disabled icons in the other screens will be like below:
Screen Shot 2019-08-12 at 6 56 46 PM

To avoid this we have added this code change giving selected style along with the disabled.

When there is data, shouldForceDisableLegendItem will be false. So, it will not impact if you select/deselect.

Hope I have answered your question. Let me know if any.

Thanks,

abhijit945: Sounds good! Did I miss the unit test for this? I can see one for `aria-disabled`.

abhijit945:
Looks good to me 👍
This will be merged once [CARBON-49](https://jira3.cerner.com/browse/CARBON-49) is closed.

Issue #41, opened 2019-07-11
Navigation buttons to move left and right on a chart/graph


abhijit945:
There were three options for Panning in Carbon, each with pros and cons: 

1. Browser default scroll bar
- this is the browser default scroll bar because there is too much data to show
- would appear if we turn off Carbon resize
- there is no return button option because all data is loaded
- the consumer decides what the look back is and what “all data” means (the look back fetches more data, for example 1mo, 1 year, by consumer control)
- this isn’t carbon, isn’t really panning, this is all browser
- it would be arrow based navigation for keyboard accessibility
- would work for any graph type

2. Return/left/right button control
- Return/left/right button control
- Doesn’t load all data in advance
- Carbon will show loading while the consumer feeds more data
- consumer provides answers each interaction to if there is more data outside of the arrows. we can dither the arrow if there isn’t more data outside the view. carbon will save the state of initial load and return will take them there.
- We won’t disable resize. 
- Works for multiple graph types
- Would be tab to navigate, enter to select
- Text on the return button would require each consumer to define, or Carbon could do an icon and English-only Tooltip
- Works better with zooming buttons 

3. D3 panning
- All data must be loaded in advance
- no keyboard navigation
- no indication you can pan
- no return button option
- Would disable resize because it is a requirement from d3 to use panning
- only works for line and bar
- We would have to disable zoom out
- https://bl.ocks.org/mbostock/7ec977c95910dd026812

We cannot make our own scroll bar (browser is the only option) and we cannot have #2 or #3 together because we won’t know what a “lower step” is for each of the left/right buttons because everything is loaded. Ultimately, to be the most accessible, we will move forward with only left and right arrows to move by each lower step increments because panning from D3 offers no affordance you can click and drag. We will not have a return button at the moment because text is not well supported in this environment and there is not a good icon for return that is as universally understood as the navigation arrows. We can also dither the arrows based on knowing if there is more data to either side of the visible graph. 

They will be placed in the upper right-hand corner of the graph aligning with edge of the y-axis. 

![image](https://user-images.githubusercontent.com/4908886/61049991-88b7e900-a3ab-11e9-8512-4364c5d0e0ae.png)

![image](https://user-images.githubusercontent.com/4908886/61050027-a08f6d00-a3ab-11e9-866e-dfa1d3ae4b07.png)

![image](https://user-images.githubusercontent.com/4908886/61050034-a4bb8a80-a3ab-11e9-835f-933ab7d3ed3b.png)

[...]

PR #67, opened 2019-08-20
New: Panning property added

Prashant-Kamboj:

[] Documentation update
[] Bug fix
[X] New functionality
[] Other, please explain:

Added pan object to graph config, removed clamping if pan enabled , removed transition if pan enabled and added def to include the dateline indicator if pan enabled
**Add any screenshots**
image


src/main/js/controls/Gantt/Gantt.js
```diff
@@ -204,14 +205,14 @@ class Gantt extends Construct {
...
             );
-        createDefs(this.config, this.svg);
         createGrid(this.axis, this.scale, this.config, this.svg);
         createContentContainer(this.config, this.svg);
         createAxes(this.axis, this.scale, this.config, this.svg);
         createGanttContent(this.config, this.svg);
         if (utils.notEmpty(this.config.dateline)) {
             createDateline(this.scale, this.config, this.svg);
         }
+        createDefs(this.config, this.svg);
```
abhijit945: Based on your email: `I have changed the order in which the def is created because when the empty graph is loaded with dateline ( with panning enabled ) the createDefs is called before the createDateline so I changed the order to call the createDateline first then createDefs.`

What was the impact when panning is not used? 
What tests were breaking because of this?

src/main/js/controls/Gantt/helpers/creationHelpers.js
```diff
@@ -291,11 +291,22 @@ const getYAxisRange = (trackList) =>
  * @returns {undefined} - returns nothing
  */
 const scaleGraph = (scale, config) => {
-    scale.x = d3.time
-        .scale()
-        .domain(config.axis.x.domain)
-        .range([0, getXAxisWidth(config)])
-        .clamp(true);
+    if (
+        config.pan !== undefined &&
```
abhijit945: I feel like we talked about this before didnt we? 
We need to create a dictionary of settings that we are changing when panning is enabled. https://github.com/cerner/carbon-graphs/issues/41#issuecomment-510461080.

src/main/js/controls/Graph/helpers/helpers.js
```diff
@@ -75,18 +76,45 @@ const getAxisInfoRowLabelHeight = (config) =>
...
-    canvasSVG
-        .select(`clipPath#${config.clipPathId}`)
-        .selectAll("rect")
-        .attr("height", config.height)
-        .attr("width", getXAxisWidth(config))
-        .attr(
-            constants.Y_AXIS,
-            config.axis.x.orientation && calculateVerticalPadding(config)
-        );
+const translateDefs = (config, canvasSVG) => {
```
abhijit945: I thought we were going to create another `clipPath` for dateline. Why are you modifying the existing one?

src/main/js/helpers/axis.js
```diff
@@ -925,8 +926,7 @@ const translateAxes = (axis, scale, config, canvasSVG) => {
     }
     canvasSVG
         .select(`.${styles.axisInfoRow}`)
-        .transition()
-        .call(constants.d3Transition)
```
abhijit945: Why was this removed?

src/main/js/controls/Graph/Graph.js
```diff
@@ -247,6 +247,7 @@ class Graph extends Construct {
         ) {
             createDateline(this.scale, this.config, this.svg);
         }
+        createDefs(this.config, this.svg);
```
abhijit945: `defs` is usually one of the first elements created within an SVG. Lets not break away from that tradition.

src/main/js/controls/Graph/helpers/helpers.js
```diff
@@ -75,18 +76,45 @@ const getAxisInfoRowLabelHeight = (config) =>
  * @private
  * @param {object} config - config object derived from input JSON
  * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
- * @returns {object} d3 svg path
+ * @returns {undefined} d3 svg path
```
abhijit945: Invalid return description

src/main/js/controls/Gantt/helpers/translateHelpers.js
```diff
@@ -121,14 +122,37 @@ const getTrackLabelTransformProperty = (scale, config) => (trackName) =>
  * @private
  * @param {object} config - config object derived from input JSON
  * @param {d3.selection} canvasSVG - d3 selection node of canvas svg
- * @returns {object} d3 svg path
+ * @returns {undefined} d3 svg path
```
abhijit945: Invalid return description

src/main/js/helpers/translateUtil.js
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Makes translation zero if panning is enabled
+ *
+ * @private
+ * @param {object} config - config object derived from input JSON
+ * @returns {undefined} transcall - holds the value of transition.
+ */
+const translatePan = (config) => {
+    let transcall;
```
abhijit945: I like the idea but can we have this in a way where it is more functional rather than procedural in nature?
Lets say that I have a dictionary entry based on https://github.com/cerner/carbon-graphs/issues/41#issuecomment-510461080

```js
// In GraphConfig.js, GanttConfig.js etc.
const settingsDictionary = (config) =>
    isPanningModeEnabled(config)
        ? {
              shouldClamp: false,
              transition: {
                  duration: 0,
                  ease: "linear"
              }
          }
        : {
              shouldClamp: true,
              transition: {
                  duration: 250,
                  ease: "linear"
              }
          };

```
Now when you are using the transition, you can just do `d3Transition(settingsDictionary.transition)` instead. Change the `d3Transition` in `constants.js` function to:
```
const d3Transition = (d=250, e="linear") => (transition) => transition.duration(d).ease(e);
```

You wouldnt need this function neither `d3TransitionPan`.

src/main/js/helpers/axis.js
Prashant-Kamboj: i will correct it.

src/main/js/controls/Gantt/Gantt.js
Prashant-Kamboj: no test as such were breaking only those test cases where we are checking the order of the creation of the elements in the svg were breaking.

src/main/js/controls/Gantt/helpers/creationHelpers.js
Prashant-Kamboj: regarding dictionary of setting we didn't had any discussion only how the pan object should look like we discussed.
```
pan: {
    enabled: true,
    showControls: true
}
```
i will refer the #41 (comment) for this.

src/main/js/controls/Graph/Graph.js
Prashant-Kamboj: As in mail if we keep the call on the first place then if we load the graph without any content then since the createDateline call is below createDefs call, so we can not get the element of dateline-point at the time of createDefs call which we need to do the manipulation.

src/main/js/controls/Bar/helpers/goalLineHelpers.js
```diff
@@ -128,7 +129,7 @@ const createRegion = (scale, config, regionGroupSVG, regionList, uniqueKey) => {
     regionPath
         .exit()
         .transition()
-        .call(constants.d3Transition)
+        .call(translatePan(config))
```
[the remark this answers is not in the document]
Prashant-Kamboj: why it should take forever ?

src/main/js/controls/Bar/helpers/creationHelpers.js
```diff
@@ -66,7 +66,12 @@ const processDataPoints = (graphConfig, dataTarget) => {
         throw new Error(errors.THROW_MSG_INVALID_FORMAT_TYPE);
     }
     if (hasInvalidTicks(graphConfig.axis.x.ticks.values, dataTarget.values)) {
-        throw new Error(errors.THROW_MSG_INVALID_X_AXIS_TICK_VALUES);
+        if (
```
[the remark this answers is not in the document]
Prashant-Kamboj: I will correct it.

dev/app.js
```diff
@@ -152,6 +154,11 @@ renderSiteApp(
                     pathname: "/line/shapes-hidden",
                     content: renderLineShapesHidden,
                     title: "Shapes Hidden"
+                },
+                {
+                    pathname: "/line/Panning",
```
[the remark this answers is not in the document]
Prashant-Kamboj: i will correct it.

dev/app.js
```diff
@@ -327,6 +334,11 @@ renderSiteApp(
                     pathname: "/gantt/label-truncation",
                     content: renderGanttTruncate,
                     title: "Label Truncation"
+                },
+                {
+                    pathname: "/gantt/Panning Mode",
```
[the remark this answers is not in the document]
Prashant-Kamboj: i will correct it.

src/main/js/helpers/translateUtil.js
Prashant-Kamboj: `// In GraphConfig.js, GanttConfig.js etc.`
Will it be worth to add same peace of code in all the config files, it will be a redundant piece of code.

src/main/js/controls/Graph/helpers/helpers.js
Prashant-Kamboj: what will be correct return for this i am not sure.

src/main/js/controls/Graph/helpers/helpers.js
Prashant-Kamboj: Because i have to do the change at one place only.

If it's not proper we can go for different function.

src/main/js/helpers/translateUtil.js
abhijit945: That's ok for now, we already have a lot of duplicate code that we might want to trim later on. 
As for this piece of code, it will be in 3 different places as opposed to considerable more amount of if-else checks.

src/main/js/controls/Graph/Graph.js
abhijit945: Creation can always be done first, the attributes of `rect` can be updated when the indicator has been added.

src/main/js/controls/Graph/helpers/helpers.js
abhijit945: `@returns {undefined} - returns nothing`

src/main/js/controls/Graph/helpers/helpers.js
```diff
@@ -313,18 +341,48 @@ const updateAxesDomain = (config, input = {}) => {
...
```
abhijit945: `@returns {undefined} - returns nothing`

src/main/js/controls/Graph/Graph.js
Prashant-Kamboj: ok, then inside the createDateline I have to update the attributes of defs right?

src/main/js/helpers/translateUtil.js
Prashant-Kamboj: ok, I will do the above changes.

Prashant-Kamboj: Since transition is at many places like axis.js dateline.js etc we should have a commonplace for our logic this way it won't work.

Prashant-Kamboj: Let's have this function in a different js file in helpers?

Prashant-Kamboj:
Regarding the testing of transition how we need to test it because it's in almost all files.

abhijit945:
One approach would be to test if
* All the current tests pass
* When in panning mode, correct duration and transition type is provided to the d3 transition function.

docs/controls/Gantt.md
```diff
@@ -377,6 +378,9 @@ var input = {
             value: new Date(2018, 5, 1).toISOString()
         }
     ],
+    pan: {
```
abhijit945: Since this is disabled by default. Lets not include it here.

docs/controls/Gantt.md
```diff
@@ -80,6 +80,7 @@ ganttInstance.loadContent(/* Add "dataArray" JSON, shown below for example */);
 | showActionLegend | boolean  | true                                           | Toggle to show graph legend                                                          |
 | padding          | object   | `{ top: 10, bottom: 5, left: 100, right: 50 }` | Refer [Padding](Padding.md)                                                          |
 | clickPassThrough | object   | null                                           | Refer [Pass Through](#Pass-Through)                                                  |
+| pan              | object   | {}                                             | Set enabled to true inside pan to enable panning mode                                |
```
abhijit945: Add a section for `Pan` property and explain what is has. 
Since this is a new property that is common for Graph, Gantt and Timeline. Lets create a new markdown for this similar to `Padding.md` and link it rather than duplicating it.

abhijit945: You can just say `Refer [Panning](Panning.md)`

src/main/js/controls/Bar/Bar.js
```diff
@@ -191,7 +193,8 @@ class Bar extends GraphContent {
             this.ordinalScale,
             graph.svg,
             this.dataTarget,
-            graph.config
+            graph.config,
+            settingsDictionary(graph.config).transition
```
abhijit945: This needs to be a 1 time call in GraphConfig since you wont expect any different output no matter where you call. You can store the output as part of the config [here](https://github.com/cerner/carbon-graphs/blob/a47fc48079b8970f8f5e538ff402cb2b79107b45/src/main/js/controls/Graph/GraphConfig.js#L85). 
```javascript
config.settingsDictionary = settingsDictionary(input);
```
Now you will be able to directly reference `graph.config.settingsDictionary.shouldClamp` or `graph.config.settingsDictionary.transition` wherever you need (in graph or any content).

src/main/js/controls/Bar/helpers/goalLineHelpers.js
```diff
@@ -128,7 +129,7 @@ const createRegion = (scale, config, regionGroupSVG, regionList, uniqueKey) => {
     regionPath
         .exit()
         .transition()
-        .call(constants.d3Transition)
+        .call(constants.d3Transition(settingsDictionary(config).transition))
```
abhijit945: `config.settingsDictionary.transition`

src/main/js/controls/Bar/helpers/goalLineHelpers.js
```diff
@@ -139,12 +140,13 @@ const createRegion = (scale, config, regionGroupSVG, regionList, uniqueKey) => {
  * @param {object} scale - d3 scale taking into account the input parameters
  * @param {object} config - config object derived from input JSON
  * @param {object} regionGroupSVG - d3 object of region group svg
+ * @param {object} transition - gets transition based on pannig mode is enabled or not
  * @returns {undefined} - returns nothing
  */
-const translateRegion = (scale, config, regionGroupSVG) => {
+const translateRegion = (scale, config, regionGroupSVG, transition) => {
```
abhijit945: This parameter is not necessary since you will have the required property `config.settingsDictionary.transition` within config.

src/main/js/controls/Gantt/Gantt.js
```diff
@@ -210,7 +211,12 @@ class Gantt extends Construct {
         createAxes(this.axis, this.scale, this.config, this.svg);
         createGanttContent(this.config, this.svg);
         if (utils.notEmpty(this.config.dateline)) {
-            createDateline(this.scale, this.config, this.svg);
+            createDateline(
+                this.scale,
+                this.config,
+                this.svg,
+                settingsDictionary(this.config).transition
```
abhijit945: Same rules as Graph.js apply here. You dont need a separate parameter if you are using properties within `this.config`

src/main/js/controls/Gantt/GanttConfig.js
```diff
@@ -111,6 +114,42 @@ const validateEventData = (content) => {
...
     }
 };
+/**
+ * Checks if panning is enabled or not
+ *
+ * @private
+ * @param {object} config - config object used by the graph.
+ * @returns {boolean} returns true of panning enabled else false.
+ */
+export const isPanningModeEnabled = (config) => {
```
abhijit945: How about just
```javascript
export const isPanningModeEnabled = (config) => utils.isDefined(config.pan) && config.pan.enabled === true;
```
?

src/main/js/controls/Gantt/GanttConfig.js
```diff
@@ -111,6 +114,42 @@ const validateEventData = (content) => {
...
+    isPanningModeEnabled(config)
+        ? {
+              shouldClamp: false,
+              transition: {
+                  duration: 0,
+                  ease: "linear"
+              }
+          }
+        : {
+              shouldClamp: true,
```
abhijit945: Do we want to have a property here to create dateline def? Only create the def and assign URL if this property is enabled?

src/main/js/controls/Gantt/helpers/creationHelpers.js
```diff
@@ -310,18 +311,36 @@ const scaleGraph = (scale, config) => {
...
         .append("clipPath")
         .attr("id", config.clipPathId)
         .append("rect")
         .attr(constants.X_AXIS, getXAxisXPosition(config))
         .attr(constants.Y_AXIS, getYAxisYPosition(config))
         .attr("width", getXAxisWidth(config))
         .attr("height", getYAxisHeight(config));
+
+    if (
+        config.pan !== undefined &&
```
abhijit945: Ah you are doing it. 💯 
This is what I meant, since we already have the `settingsDictionary` within config, we could just listen to what that new property would say instead of rechecking if the panning is enabled.

src/main/js/controls/Gantt/helpers/translateHelpers.js
```diff
@@ -405,12 +433,15 @@ const translateDataPoints = (scale, config, trackPath) => {
  * @private
  * @param {object} scale - d3 scale for Graph
  * @param {Selection} path - d3 object of track
+ * @param {object} graphConfig - graph config needed for panning feature
  * @returns {undefined} - returns nothing
  */
-const translateTaskBar = (scale, path) =>
+const translateTaskBar = (scale, path, graphConfig) =>
```
abhijit945: `config`. Let's not confuse this with Graph API's config.

src/main/js/controls/Gantt/helpers/translateHelpers.js
```diff
@@ -441,12 +472,15 @@ const translateTaskBar = (scale, path) =>
  * @private
  * @param {object} scale - d3 scale for Graph
  * @param {Selection} path - d3 object of track
+ * @param {object} graphConfig - graph config needed for panning feature
  * @returns {undefined} - returns nothing
  */
-const translateActivityBar = (scale, path) =>
+const translateActivityBar = (scale, path, graphConfig) =>
```
abhijit945: `config`

src/main/js/controls/Graph/GraphConfig.js
```diff
@@ -164,7 +168,42 @@ export const validateContent = (content, input) => {
...
 };
-
+/**
+ * Checks if panning is enabled or not
+ *
+ * @private
+ * @param {object} config - config object used by the graph.
+ * @returns {boolean} returns true of panning enabled else false.
+ */
+export const isPanningModeEnabled = (config) => {
```
abhijit945: Same comment as mentioned in `GanttConfig`. If you want we could even move this function to BaseConfig and have Gantt, Graph and Timeline use from there.

src/main/js/controls/Graph/GraphConfig.js
```diff
@@ -164,7 +168,42 @@ export const validateContent = (content, input) => {
...
+    return false;
+};
+/**
+ * Used to set the clamp and transition when panning is enabled or not.
+ *
+ * @private
+ * @param {object} config - config object used by the graph.
+ * @returns {undefined} returns nothing
+ */
+export const settingsDictionary = (config) =>
```
abhijit945: While we are considering moving `isPanningModeEnabled` function to `BaseConfig` we **cant** move this function. 
If you are considering including the dateline property here. (Timeline doesnt have a dateline).

src/main/js/controls/Graph/helpers/helpers.js
abhijit945: No, this is ok

src/main/js/controls/Pie/PieConfig.js
```diff
@@ -21,6 +21,42 @@ export const processInput = (input, config) => {
...
     return config;
 };
+/**
+ * Checks if panning is enabled or not
+ *
+ * @private
+ * @param {object} config - config object used by the graph.
+ * @returns {boolean} returns true of panning enabled else false.
+ */
+export const isPanningModeEnabled = (config) => {
```
abhijit945: Why does Pie chart need panning?

src/main/js/controls/Timeline/TimelineConfig.js
```diff
@@ -48,6 +48,42 @@ export const processInput = (input, config) => {
...
+ *
+ * @private
+ * @param {object} config - config object used by the graph.
+ * @returns {undefined} returns nothing
+ */
+export const settingsDictionary = (config) =>
+    isPanningModeEnabled(config)
+        ? {
+              shouldClamp: false,
+              transition: {
```
abhijit945: Lets create a constant for these 2 and add it to constants.js so that everyone can reference it. 
```js
CONSTANTS.D3_TRANSITION_PROPERTIES_ENABLED = {
	duration: 250,
	ease: "linear"
}
CONSTANTS.D3_TRANSITION_PROPERTIES_DISABLED = {
	duration: 0,
	ease: ""
}
```

src/main/js/controls/Timeline/TimelineConfig.js
```diff
@@ -48,6 +48,42 @@ export const processInput = (input, config) => {
...
+                  duration: 0,
+                  ease: "linear"
```
abhijit945: You can remove the ease by providing `""`

src/main/js/helpers/dateline.js
```diff
@@ -144,7 +164,39 @@ const createDateline = (scale, config, canvasSVG) => {
...
+        const datelineIndicatorHeight = Math.floor(
+            Math.max(...shapeHeightArr) / 2
+        );
+        canvasSVG
+            .select(`clipPath#${config.datelineClipPathId}`)
+            .selectAll("rect")
+            .attr(
+                "height",
+                getYAxisHeight(config) === 0
+                    ? datelineIndicatorHeight + 20
```
abhijit945: Why the arbitrary numbers? 20, 4?

src/main/js/helpers/translateUtil.js
```diff
@@ -0,0 +1,20 @@
+import constants from "./constants";
+
+/**
```
abhijit945: We can safely remove this file?

Prashant-Kamboj: yes I will remove this file

src/main/js/helpers/dateline.js
Prashant-Kamboj: because fro gantt the defs for dateline when there is no content the height is zero so since the indicators are little above the axis they are hidden so I have attached the what is least height and y-axis I can provide so that for no content the indicators are visible since dateline function is a common one.

src/main/js/controls/Pie/PieConfig.js
Prashant-Kamboj: I will remove it

dev/examples/controls/line.js
```diff
@@ -1,6 +1,7 @@
 import Carbon from "../../../src/main/js/carbon";
 import utils from "../../../src/main/js/helpers/utils";
 import { getDemoData } from "../data";
+import { createPanningControls } from "./panHelpers";
```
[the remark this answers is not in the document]
Prashant-Kamboj: Are you saying that are we still getting the issue on the site?

dev/examples/controls/line.js
```diff
@@ -444,3 +445,32 @@ export const renderLineCustomPadding = (id) => {
...
     return lineDefault;
 };
+export const renderLineWithPanning = (id) => {
+    let graph;
+    const axisData = utils.deepClone(
+        getDemoData(`#${id}`, "LINE_TIMESERIES_DATELINE")
+    );
+    axisData.pan = {
+        enabled: true,
+        showControl: true
```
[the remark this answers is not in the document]
Prashant-Kamboj: yes, I will remove it.

docs/controls/Gantt.md
Prashant-Kamboj: I will remove it

src/main/js/controls/Graph/GraphConfig.js
Prashant-Kamboj: As per below comment we are not moving it to baseConfig

src/main/js/controls/Gantt/GanttConfig.js
Prashant-Kamboj: ok then what name should be given to that property like "createDatelineDef"

src/main/js/controls/Gantt/helpers/translateHelpers.js
Prashant-Kamboj: I will correct it

dev/examples/controls/line.js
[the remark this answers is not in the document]
abhijit945: Didnt you say that there would be console errors in site when you use navigation buttons and resize in any other page?

src/main/js/controls/Gantt/GanttConfig.js
abhijit945: How about `shouldCreateDatelineDefs: true/false`?

src/main/js/controls/Pie/PieConfig.js
abhijit945: Why not just pass the constant with enabled transition `D3_TRANSITION_PROPERTIES_ENABLED`, when you are calling `constants.d3Transition`?

src/main/js/controls/Graph/GraphConfig.js
abhijit945: Sorry for the misunderstanding, I was saying that `isPanningModeEnabled` function can be moved but `settingsDictionary` cannot be.

src/main/js/helpers/dateline.js
abhijit945: I mean if this is Gantt specific, then doing this will affect Graph construct as well. Is that what you want?

abhijit945:
I would also move the `Panning` from [here](https://github.com/cerner/carbon-graphs#potential-features) to [here](https://github.com/cerner/carbon-graphs#other-highlights) 😃

src/main/js/helpers/dateline.js
Prashant-Kamboj: No it will not affect the graph construct it will work only for gantt because for graph the y-axis will never be zero without any content but for gantt graph it will become zero for without any content.

src/main/js/controls/Gantt/GanttConfig.js
Prashant-Kamboj: ok I will add this property

dev/examples/controls/line.js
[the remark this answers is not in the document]
Prashant-Kamboj: No, it's still there as per discussion we would be logging it as an issue in the site which I can log only when it will be released.

src/main/js/helpers/dateline.js
abhijit945: I am just concerned adding magic numbers like this would affect adversely when other parameters like padding are changed. We are already seeing issues with having arbitrary numbers everywhere. 
I am not sure why we are even showing the dateline when there is no content in Gantt. Since unlike Graph Construct, the content in Gantt is what makes the height of the graph. 
* Can you add some screenshots for Gantt with no data and with dateline (Without panning)? 
* Also screenshots with Gantt before and after adding the 20 and 4?

src/main/js/controls/Bar/Bar.js
Prashant-Kamboj: I will make the changes

src/main/js/controls/Bar/helpers/goalLineHelpers.js
Prashant-Kamboj: Yes, I will correct it.

src/main/js/controls/Bar/helpers/goalLineHelpers.js
Prashant-Kamboj: yes, I will remove it after adding settingDictionary in config

src/main/js/controls/Pie/PieConfig.js
Prashant-Kamboj: Do you mean to hardcode "D3_TRANSITION_PROPERTIES_ENABLED" for pie at all the places where we are calling transition?

src/main/js/controls/Graph/GraphConfig.js
Prashant-Kamboj: ok sure I will move the isPanningModeEnabled function ti BaseConfig

src/main/js/controls/Gantt/helpers/translateHelpers.js
Prashant-Kamboj: I will correct it

src/main/js/controls/Gantt/helpers/creationHelpers.js
Prashant-Kamboj: I will make the change

src/main/js/controls/Timeline/TimelineConfig.js
Prashant-Kamboj: ok I will make the change.

src/main/js/controls/Timeline/TimelineConfig.js
Prashant-Kamboj: ok will add the constants

docs/controls/Gantt.md
Prashant-Kamboj: ok, I will make a new Panning .md

src/main/js/helpers/dateline.js
Prashant-Kamboj: image

src/main/js/controls/Gantt/helpers/translateHelpers.js
```diff
@@ -121,14 +121,43 @@ const getTrackLabelTransformProperty = (scale, config) => (trackName) =>
...
+        const datelineIndicatorHeight = Math.floor(
+            Math.max(...shapeHeightArr) / 2
+        );
+        canvasSVG
+            .select(`clipPath#${config.datelineClipPathId}`)
+            .selectAll("rect")
+            .attr(
+                "height",
+                config.height === 0
+                    ? datelineIndicatorHeight + 5
```
abhijit945: This condition is no longer needed is it? We discussed that https://github.com/cerner/carbon-graphs/issues/88 would take care of height not being 0.

src/main/js/helpers/dateline.js
```diff
@@ -144,7 +164,32 @@ const createDateline = (scale, config, canvasSVG) => {
...
-    translateDateline(scale, config, canvasSVG, getYAxisYPosition);
+    translateDateline(scale, config, canvasSVG, getYAxisYPosition, transition);
+
+    if (
+        config.settingsDictionary.shouldCreateDatelineDefs &&
+        config.dateline.length > 0
+    ) {
+        const shapeHeightArr = [];
+        const shape = d3.selectAll(`.${styles.datelinePoint}`);
+        shape[0].forEach((element) => {
```
abhijit945: I am wondering if we can move this in `dateline.js`. We are essentially duplicating in Gantt, Graph and Timeline with some minor to no changes, arent we?

src/test/unit/controls/Gantt/Gantt-spec.js
```diff
@@ -2491,4 +2492,79 @@ describe("Gantt", () => {
             expect(gantt.resizeHandler).toBeNull();
         });
     });
+    describe("When pan is enabled", () => {
```
abhijit945: `describe("Panning", () => {`
We are checking stuff in here that are applicable when panning is not enabled as well.

src/test/unit/controls/Gantt/Gantt-spec.js
```diff
@@ -2491,4 +2492,79 @@ describe("Gantt", () => {
...
+                const translate = getSVGAnimatedTransformList(
+                    datelineGroupElement.getAttribute("transform")
+                ).translate;
+                expect(toNumber(translate[0], 10)).toBeCloseTo(106);
+                expect(toNumber(translate[1], 10)).toBeCloseTo(5);
+                done();
+            }, 10);
+        });
+    });
+    describe("When pan is disabled", () => {
```
abhijit945: Move this inside. 
```
describe Panning
    describe When enabled
    describe When disabled
```

src/test/unit/controls/Gantt/Gantt-spec.js
```diff
@@ -2491,4 +2492,79 @@ describe("Gantt", () => {
...
+            const datelineGroupElement = fetchElementByClass(
+                styles.datelineGroup
+            );
+            expect(d3.select(datelineGroupElement).datum().value).toBe(
+                datelineAlt.value
+            );
+            expect(datelineGroupElement.getAttribute("aria-selected")).toBe(
+                "false"
+            );
+            delay(() => {
```
abhijit945: Great work! 💯

src/test/unit/controls/Gantt/Gantt-spec.js
```diff
@@ -2491,4 +2492,79 @@ describe("Gantt", () => {
...
+        });
+    });
+    describe("When pan is disabled", () => {
+        beforeEach(() => {
+            const axisData = utils.deepClone(getAxes(axisJSON));
+            axisData.dateline = datelineJSON;
+            axisData.pan = { enabled: false };
+            gantt = new Gantt(axisData);
+        });
+        it("Check if clamp is false if pan is enabled", () => {
```
abhijit945: Might want to add a test case that the panning is disabled when consumer doesnt provide anything? pan property is `undefined`?

Prashant-Kamboj: Sure I will add the test case.

dev/examples/controls/bar.js
```diff
@@ -364,3 +365,67 @@ export const renderStackedBarAxisInfoTextLabels = (id) => {
...
+        ],
+        format: "%H"
+    };
+    const graphData = {
+        key: "uid_bar_t1",
+        label: {
+            display: "Data Label t1"
+        },
+        color: Carbon.helpers.COLORS.BLUE,
+        //onClick: loadBarPopup,
```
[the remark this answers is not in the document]
Prashant-Kamboj: I will import the function and uncomment it.

dev/examples/controls/gantt.js
```diff
@@ -321,6 +345,51 @@ export const renderGanttPercentage = (id) => {
...
+            onClick: loadTrackLabelPopup
+        },
+        tasks: tasks2[0]
+    };
+    const createGraph = (axis, values) => {
+        if (graph) {
+            graph.destroy();
+        }
+        graph = Carbon.api.gantt(axis);
+        //graph.loadContent(values);
```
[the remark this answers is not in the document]
Prashant-Kamboj: I will uncomment it, I was testing how gantt is behaving after removing the magic number for without content.

src/main/js/controls/Gantt/helpers/translateHelpers.js
Prashant-Kamboj: ok sure I will remove it

src/test/unit/controls/Gantt/Gantt-spec.js
Prashant-Kamboj: I will change the description

docs/controls/Panning.md
```diff
@@ -0,0 +1,34 @@
...
+
+| Property Name | Expected | Description                        |
+| ------------- | -------- | ---------------------------------- |
+| `enabled`     | boolean  | Set to true when panning is needed |
+
+## Constraints
+
+-   If panning is not provided then enabled will be false.
+
+## Default Panning
```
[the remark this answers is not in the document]
Prashant-Kamboj: ok I will remove this

docs/controls/Panning.md
```diff
@@ -0,0 +1,34 @@
...
+
+-   By defalut enabled will be false to enable set it to true as below
+
+```javascript
```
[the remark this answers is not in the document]
Prashant-Kamboj: like this
## Usage

```javascript
var pan = {
    enabled: true
};
```

src/main/js/controls/Graph/Graph.js
```diff
@@ -247,7 +248,12 @@ class Graph extends Construct {
             utils.notEmpty(this.config.dateline) &&
             this.config.axis.x.type === AXIS_TYPE.TIME_SERIES
         ) {
-            createDateline(this.scale, this.config, this.svg);
+            createDateline(
+                this.scale,
+                this.config,
+                this.svg,
+                this.config.settingsDictionary.transition
```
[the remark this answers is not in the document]
Prashant-Kamboj: yes right I will make this change

docs/controls/Panning.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Root
+
+#### Required
+
+| Property Name | Expected | Description                        |
+| ------------- | -------- | ---------------------------------- |
+| `enabled`     | boolean  | Set to true when panning is needed |
+
+## Constraints
```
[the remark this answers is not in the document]
Prashant-Kamboj: We can keep this I think

docs/controls/Panning.md
```diff
@@ -0,0 +1,34 @@
...
+        -   [Root](#root)
+            -   [Required](#required)
+    -   [Constraints](#constraints)
+    -   [Default Panning](#default-panning)
+
+## JSON Properties
+
+### Root
+
+#### Required
```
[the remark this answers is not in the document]
Prashant-Kamboj: Ok, I will change Required to Optional

dev/examples/controls/pairedResult.js
```diff
@@ -320,3 +321,86 @@ export const renderPairedResultXOrientationTop = (id) => {
...
+                    isCritical: false
+                },
+                mid: {
+                    x: "2015-12-31T20:30:00.000Z",
+                    y: 40
+                },
+                low: {
+                    x: "2015-12-31T20:30:00.000Z",
+                    y: 10,
+                    isCritical: false
```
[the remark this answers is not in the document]
Prashant-Kamboj: Yes I will remove it

src/main/js/helpers/dateline.js
Prashant-Kamboj: Since both the gantt and graph uses different way to calculate the "getYAxisYPosition" that is why we can not make the function common for both graph and gantt.

docs/controls/Panning.md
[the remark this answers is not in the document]
abhijit945: You missed the default column here.

src/main/js/controls/Gantt/helpers/translateHelpers.js
```diff
@@ -121,14 +121,43 @@ const getTrackLabelTransformProperty = (scale, config) => (trackName) =>
...
         .selectAll("rect")
         .attr("height", getYAxisHeight(config))
         .attr("width", getXAxisWidth(config));
+    if (
+        config.settingsDictionary.shouldCreateDatelineDefs &&
+        config.dateline.length > 0
+    ) {
+        const shapeHeightArr = [];
+        const shape = d3.selectAll(`.${styles.datelinePoint}`);
+        shape[0].forEach((element) => {
```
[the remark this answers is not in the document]
abhijit945: This is pending

src/main/js/controls/Graph/Graph.js
```diff
@@ -318,6 +319,9 @@ class Graph extends Construct {
         if (utils.notEmpty(content.config.values)) {
             removeNoDataView(this.svg);
         }
+        if (utils.notEmpty(content.config.values)) {
```
abhijit945: Why is this duplicated?

src/main/js/controls/Graph/helpers/helpers.js
```diff
@@ -87,6 +87,34 @@ const translateDefs = (config, canvasSVG) =>
...
+
+    if (
+        config.pan !== undefined &&
+        utils.isBoolean(config.pan.enabled) &&
+        config.pan.enabled &&
+        config.dateline.length > 0
+    ) {
+        const shapeHeightArr = [];
+        const shape = d3.selectAll(`.${styles.datelinePoint}`);
+        shape[0].forEach((element) => {
```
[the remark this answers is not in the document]
abhijit945: Do this:

```js
d3.selectAll(`.${styles.datelinePoint}`).each(function() {
    const shapeHeight = this.getBBox().height;
    shapeHeightArr.push(shapeHeight);
});
```

src/main/js/controls/Graph/helpers/helpers.js
```diff
@@ -87,6 +87,34 @@ const translateDefs = (config, canvasSVG) =>
...
+            const shapeHeight = element.getBBox().height;
```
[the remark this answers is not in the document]
abhijit945: Pending.

src/main/js/controls/Graph/helpers/helpers.js
```diff
@@ -87,6 +87,32 @@ const translateDefs = (config, canvasSVG) =>
...
+            shapeHeightArr.push(shapeHeight);
+        });
+        const datelineIndicatorHeight = Math.floor(
+            Math.max(...shapeHeightArr) / 2
```
abhijit945: Did I miss the unit test for this?

src/main/js/controls/Graph/helpers/helpers.js
```diff
@@ -87,6 +87,32 @@ const translateDefs = (config, canvasSVG) =>
...
         );
+
+    if (
+        config.settingsDictionary.shouldCreateDatelineDefs &&
+        config.dateline.length > 0
+    ) {
+        const shapeHeightArr = [];
+        const shape = d3.selectAll(`.${styles.datelinePoint}`);
+        shape[0].forEach((element) => {
+            const shapeHeight = element.getBBox().height;
```
abhijit945: Doing a `getBBox` will cause a repaint on the page. Since this is called on every resize, it would lead to repaint everytime. 
Can we do this just once and then reuse? The values are not going to change are they?

src/main/js/core/BaseConfig/helper.js
```diff
@@ -109,3 +118,16 @@ export const parseTypedValue = (x, xAxisType) =>
...
         : utils.getNumber(x);
+/**
+ * Checks if panning is enabled or not
+ *
+ * @private
+ * @param {object} config - config object used by the graph.
+ * @returns {boolean} returns true of panning enabled else false.
+ */
+export const isPanningModeEnabled = (config) => {
+    if (config.pan !== undefined && config.pan.enabled) {
```
abhijit945: From a previous pending comment
```js
export const isPanningModeEnabled = (config) => utils.isDefined(config.pan) && config.pan.enabled === true;
```

src/main/js/helpers/dateline.js
```diff
@@ -99,13 +99,25 @@ const datelineClickHandler = (value, target) => {
  * @private
  * @param {object} scale - d3 scale taking into account the input parameters
  * @param {object} config - config object derived from input JSON
- * @param {Array} canvasSVG - d3 object of canvas svg
+ * @param {Array} canvasSVG - d3 object of canvas sv
```
abhijit945: typo

src/main/js/helpers/datetimeBuckets.js
```diff
@@ -204,9 +226,10 @@ const translateVGridDatetimeBuckets = (
  * @param {Array} values - list of all tick values.
  * @param {string} height - Y axis height.
  * @param {string} orientation - X axis orientation.
+ * @param {object} transit - gets transition based on pannig mode is enabled or not
  * @returns {object} - config with tick values and height.
  */
-const createDatetimeBucketConfig = (values, height, orientation) => ({
+const createDatetimeBucketConfig = (values, height, orientation, transit) => ({
```
abhijit945: Avoid unnecessary abbreviation. 
`transition`

```js
height: `${height}`,
    settingsDictionary: {
        transition
    }
```

src/main/js/controls/Graph/Graph.js
Prashant-Kamboj: I have removed it I got added by mistake

dev/app.js
```diff
@@ -623,6 +624,21 @@ renderSiteApp(
...
+            children: [
+                {
+                    pathname: "/panning/line",
+                    children: [
+                        {
+                            pathname: "/panning/line/simple",
+                            content: renderLineWithPanning,
+                            title: "Simple"
+                        }
+                    ]
```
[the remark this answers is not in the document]
Prashant-Kamboj: I have added an example for all graph types

src/main/js/core/BaseConfig/helper.js
Prashant-Kamboj: I will make the change

src/main/js/helpers/dateline.js
Prashant-Kamboj: I will correct it

src/main/js/helpers/datetimeBuckets.js
Prashant-Kamboj: I will correct it

src/main/js/controls/Gantt/helpers/translateHelpers.js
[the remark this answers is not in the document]
Prashant-Kamboj: I will do this

src/main/js/controls/Graph/helpers/helpers.js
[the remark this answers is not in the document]
Prashant-Kamboj: I will do this

src/main/js/controls/Graph/helpers/helpers.js
Prashant-Kamboj: I will place the calculation of the height of dateline indicator in dateline.js and import it where required

src/main/js/controls/Graph/helpers/helpers.js
Prashant-Kamboj: I will add the test for this

src/main/js/controls/Gantt/helpers/translateHelpers.js
```diff
@@ -121,14 +124,32 @@ const getTrackLabelTransformProperty = (scale, config) => (trackName) =>
...
         .select(`clipPath#${config.clipPathId}`)
         .selectAll("rect")
         .attr("height", getYAxisHeight(config))
         .attr("width", getXAxisWidth(config));
+    if (
+        config.settingsDictionary.shouldCreateDatelineDefs &&
+        config.dateline.length > 0
+    ) {
+        const datelineIndicatorHeight = Math.floor(
+            getDatelineIndicatorHeight() / 2
```
abhijit945: Although this is good for maintaining code. This wont help us with the performance. Let me describe why.
* Add a console.log and try resizing the page, notice how many times this function is called? 
* Calling getBoundingClientRect and getBBox essentially causes a force page reflow. 
* Calling it in a resize event is not advised since it would cause layout thrashing
More information here: https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing

But it is also true that during panning there maybe another dateline with an indicator that has a different height. So I dont see a way out of this. 
Can you add this comment in the JSDocs:
```
* @description
 * Calling getDatelineIndicatorHeight() will trigger a page reflow and resizing the page might cause Layout Thrashing.
 * We understand this and deem it necessary to calculate the indicator height when a new dataset/set of contents are loaded during Panning.
 * Furthermore, this function is called only when panning is enabled and there is a dateline defs element is present.
```

abhijit945:
@Prashant-Kamboj One thing I wanted to mention. In the graph when we dont have a data in the data-set range we have the ability to show `No Data` https://engineering.cerner.com/carbon-graphs/#/line/no-data

I was wondering if we can do this in the site for our panning example?

src/main/js/controls/Gantt/helpers/translateHelpers.js
Prashant-Kamboj: ok sure I will add comment

PR #76, opened 2019-09-11
Chore: Update concept goal-line vs region-line (fixes #38)

### LINKED ISSUE #38: Rename Goal Line to Region Line
Goal Lines are exclusive to Bar Graphs. 

To prevent confusion with Regions within Line Graph etc. when Start and End are the same. Rename this to Region Line is necessary. 
image

The confusion would also compound when Bar Graphs get their own Region per issue https://github.com/cerner/carbon-graphs/issues/12

abhijit945:

[x] Documentation update
[] Bug fix
[] New functionality
[] Other, please explain:

There wasn't a clear distinction between goal lines for Line graph and Bar graph which leads to confusion. 

Line Graph 
image

Bar Graph

image

We have plans to include Regions to Bar Graph as well at which point we will have a single line for the entire graph. 
> `Regions` in Bar Graph are conceptually different than other graph types, since `goal lines` are **per bar** as opposed to `regions lines` which are **per data-set**.


Dinesh94Singh:
In the screenshot, do we want to make it Region Line instead of Goal-line for Bar Graphs?

Dinesh94Singh:
> In the screenshot, do we want to make it Region Line instead of Goal-line for Bar Graphs?

nvm, region line are per data-set.

Issue #32, opened 2019-07-11
Axis overlap for month with 31 days


abhijit945:
**Describe the bug**
When there is date transition through 31st of a month to the 1st of the next month, then the x-axis tick values deviate from the expected behavior.

![image](https://user-images.githubusercontent.com/4908886/61050420-9326b280-a3ac-11e9-933f-54f28741bce2.png)

In the above screenshot, the two adjacent dates (tick values) are displayed and is also overlappping whereas the expected behavior should same as the one for month with 30 days as shown in the screenshot below
![image](https://user-images.githubusercontent.com/4908886/61050434-9d48b100-a3ac-11e9-99ed-261eb933aa87.png)

Noticed this behavior in the follwing graph types:
1. Line graph 
2. Paired Result Graph

This is a problem in smaller viewport sizes only.

nramamurth:
@abhijit945 

I am curious here. Is it theoretically valid to have different intervals (Mar 31 - Apr 1 being one day, and the rest being 2 days) on an axis in a Line Graph?

abhijit945:
Theoretically sure (If I am not misunderstanding you). Carbon would need to define it as an ordinal scale axis instead of timeseries (i.e. consumers will be free to define axis values as a string, so that it will be possible to define `Jan - Mar`, `Apr - Jun` etc. ). 
One such example is https://bl.ocks.org/pstuffa/c37d059efa1501d33913d63ffe3585db

We also have an enhacement issue logged for this type of axis: https://github.com/cerner/carbon-graphs/issues/11

Issue #22, opened 2019-07-11
Padding property doesn't take space into account when hasCustomPadding property is true


abhijit945:
**Describe the bug**
For bar graph, when the padding prop is provided(no matter what the value is), the Y-axis label becomes closer to Y-axis. Looks like the default white space between label and Y-axis is gone once you provided a customized padding prop.

When no padding props provided by the consumer: 
![image](https://user-images.githubusercontent.com/4908886/61050732-4d1e1e80-a3ad-11e9-899d-2cb6db389c26.png)

When provided costomized padding as `{left: 30, right: 50}`:
![image](https://user-images.githubusercontent.com/4908886/61050747-50b1a580-a3ad-11e9-9505-da0e131e01bf.png)

**To Reproduce**

1. provide the padding props in code like 
```
    axisData.padding = {
        right: 50
    };
```
or 
```
    axisData.padding = {
        left: 30,
        right: 50
    };
```

**Expected behavior**
Expected to keep the white space between Y-axis label and Y-axis when padding prop is provided.

![image](https://user-images.githubusercontent.com/4908886/61050757-55765980-a3ad-11e9-82b6-d41607ffcf9b.png)

**Screenshots**
see above.

**Desktop (please complete the following information):**
-   OS: mac OS
-   Browser: Chrome

**Additional context**

jeffmerten:
We're seeing an issue where the padding right attribute not being honored for the timeline. Not sure if this padding issue will include the resolution for that or if it would be scoped in another fix.

Santosh086:
@abhijit945 / @ryanthemanuel - Any thoughts for this issue on when it will be fixed?

Santosh086:
@abhijit945 / @ryanthemanuel - Any thoughts on this issue when it will be fixed?

abhijit945:
@pranav300 is working on the changes as of this moment. We are trying to figure out the conditions we are trying to solve along with the best possible method to achieve it passively. Since this is core to the functioning of each graph we are being very cautious. 

It is our top priority as of now. Any progress will be reflected as part of PR https://github.com/cerner/carbon-graphs/pull/83.

PR #80, opened 2019-09-18
Fix: x-axis ticks from overlapping (fixes: #32)

pranav300:

- [ ] Documentation update
- [ ] Bug fix
- [ ] New functionality
- [x]  Changes an existing functionality
- [ ] Other, please explain:

I have changed `Max_Tick_variance`, So that x-axis ticks do not overlap on each other. This is a temporary fix. This would work only on dynamic ticks, but the overlapping depends upon the size of ticks.
I will be working on issues #22 and #9 to resolve this.


abhijit945:
Can you add a screenshot with the X Axis datetime range having months instead and show the fix is valid? Thanks.

pranav300:
This is the closest 31st and 1st get.

Screenshot 2019-09-18 at 7 40 39 PM

abhijit945:
You need to also add the term `(fixes #32)` in the commit message to actually close the issue.

PR #83, opened 2019-09-18
Fix: Updated axis.js to use customPadding as constantPadding (fixes #22)

### LINKED ISSUE #22: Padding property doesn't take space into account when hasCustomPadding property is true

pranav300:

- [x] Bug fix
- [ ] Documentation update
- [ ] New functionality
- [ ] Changes an existing functionality
- [ ] Other, please explain:

**Add any screenshots**

Before

before

After

after


src/main/js/helpers/axis.js
```diff
@@ -683,16 +677,9 @@ const getYAxisWidth = (id, config) => {
...
-    if (config.padding.hasCustomPadding) {
-        return config.padding.right;
-    }
-    return (
-        (hasY2Axis(config.axis)
-            ? getYAxisWidth(constants.Y2_AXIS, config)
-            : 20) + config.padding.right
-    );
-};
+const getY2AxisWidth = (config) =>
```
abhijit945: Changes look great in Graph construct but for Gantt construct its still a bit broken. I bet that's because we are using `hasCustomPadding` property there as well. 

With adding the property as 
```
top: 10,
bottom: 5,
left: 30,
right: 50
```
**Gantt**
Gantt with `padding` property used:
image

Gantt without `padding` property used:
image

src/main/js/helpers/axis.js
```diff
@@ -714,9 +701,8 @@ const isXAxisOrientationTop = (xAxisOrientation) =>
  */
 const calculateAxesSize = (config) => {
     config.axisSizes = {};
-    config.axisSizes.y = config.padding.hasCustomPadding
-        ? getYAxisWidth(constants.Y_AXIS, config)
-        : getYAxisWidth(constants.Y_AXIS, config) + config.padding.left;
+    config.axisSizes.y =
```
abhijit945: @Dinesh94Singh Why did we need the `hasCustomPadding` flag? I am looking for some background. 
Can we remove it altogether from all the constructs (Graph, Gantt and Timeline)?

abhijit945:
When using the default padding values using padding property, it looks good on Timeline

image

vs 

image

Dinesh94Singh:
Curious to know why custom padding is getting removed and introducing constant padding. I thought we landed upon removing the constant padding altogether and do the calculations based on custom padding field itself.

https://github.cerner.com/Geneva/carbon-graphs/issues/155

src/main/js/helpers/axis.js
Dinesh94Singh: it is to determine if the customPadding is given as input.

inside getPadding function, we return the padding object from constants or default few of the fields to constants when not all fields are given as part of custom padding.

Since config.padding is assigned to the return value of getPadding, there is no way to distinguish if the padding is a custom padding or constant padding.

src/main/js/helpers/axis.js
```diff
@@ -577,9 +577,6 @@ const getYAxisHeight = (config) => config.height;
  * @returns {number} Height of the X Axis ticks, labels and numbers/datetimes
  */
 const getXAxisHeight = (config) => {
-    if (config.padding.hasCustomPadding) {
-        return config.padding.bottom;
```
Dinesh94Singh: If I remember right - removing this might break Gantt with customPadding set to all 0's. Can you post a screenshot of what happens for Gantt having 0 custom padding

pranav300: Screenshot 2019-09-20 at 11 22 45 AM

It doesn't break the code. But I am not able to understand that why are using `tracklabel` to define padding `left` both for `custom` and `constant` [GanttConfig.js](https://github.com/cerner/carbon-graphs/blob/master/src/main/js/controls/Gantt/GanttConfig.js#L133)

Dinesh94Singh: The reason might be because, track labels are displayed in the left and the length of the trackLabel defines where to start the Gantt (since we wouldn't want our labels to get cut off - giving arbitrary values).

pranav300:
I am closing this PR. And I have raised a new PR https://github.com/cerner/carbon-graphs/pull/97 to resolve the issue.

PR #87, opened 2019-09-21
Update: Added support provide key for unload Graph contents (fixes #86)

### LINKED ISSUE #86: Graph construct should only need key and label for unloading
**Is your feature request related to a problem? Please describe.**
Currently, Pie, Gantt and Timeline constructs only require key and label for unloading the content. But Graph construct requires the exact content instance to be passed for `unloadContent`. 

**Describe the solution you'd like**
This needs to be made consistent with other graph constructs. 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

abhijit945:

-   [ ] Feature
-   [ ] Bug fix
-   [ ] Dependency updates
-   [x] Documentation updates
-   [ ] Build related changes
-   [x] Changes an existing functionality

-   [ ] Yes
-   [x] No

There is an inconsistency in Graph Construct compared to other constructs such as Gantt, Timeline or Pie. 
For unloading a content, in Line, Paired Result and Bar you need to provide the same exact instance as the one loaded. Whereas its only enough if the consumer knows the unique id of that content. 

**Add any screenshots**


src/main/js/controls/Graph/Graph.js
```diff
@@ -327,17 +329,21 @@ class Graph extends Construct {
...
-     * @param {object} content - Graph content to be removed
+     * Input can be either a GraphContent instance or
+     * just an object containing a `key` of the content to be removed
+     *
+     * @param {object} input - Graph content to be removed
      * @returns {Graph} - Graph instance
      */
-    unloadContent(content) {
-        const index = this.content.indexOf(content);
+    unloadContent(input) {
```
Dinesh94Singh: Is it only for graphs ? Didn't Gantt have a similar requirement ??

abhijit945: For Gantt, Timeline and Pie it is enough to provide key and label for unloading. 
For Graph however, we are currently requiring consumer to provide the entire object. 

Consider this, 
```
var graph = new Graph();
var line1 = new Line(input);
graph.loadContent(line1);
```
For unloading you need to provide the same exact object. 
```
graph.unloadContent(line1);
```
This means that if I were to do `newLine1` with exact same input. 
```
var newLine1 = new Line(input);
graph.unloadContent(newLine1);
```
it wouldn't work , since Objects in JavaScript are compared by reference and not value.

docs/core/Lifecycle.md
```diff
@@ -0,0 +1,102 @@
...
+-   Legend
+
+Construct exposes following functions to handle the graph or its contents:
+
+-   `generate` - Called internally when instantiating the graph
+-   `loadContent` - Load a dataset in the graph
+-   `unloadContent` - Remove a dataset from the graph
+-   `resize` - Resizes the graph based on window resize event
+-   `destroy` - Destroys the graph
+
```
swatikode: If the `Construct` phase has 5 lifecycle methods then can we also indicate them along with the rest on line# 7?

docs/core/Lifecycle.md
```diff
@@ -0,0 +1,102 @@
...
+            x: "2016-04-10T12:00:00Z",
+            y: 1
+        },
+        {
+            x: "2016-11-01T12:00:00Z",
+            y: 18
+        }
+    ]
+};
+var lineGraph = Carbon.api.graph(root);
```
swatikode: I think it would be nice to say what the `root` here is. Just to make sure there wouldn't be any gaps

docs/core/Lifecycle.md
```diff
@@ -0,0 +1,102 @@
...
+lineGraph.loadContent(Carbon.api.line(data));
+```
+
+### `unloadContent(input)`
+
+To remove a dataset from graph, the consumer has to provide the correspongind unique `key` and `label` object.
```
swatikode: nit: corresponding

docs/core/Lifecycle.md
```diff
@@ -0,0 +1,102 @@
...
+
+```javascript
+lineGraph.unloadContent({
+    key: "uid_1",
+    trackLabel: {
```
swatikode: Just out of curiosity why is the `label` needed. Isn't `key` enough?

docs/core/Lifecycle.md
```diff
@@ -0,0 +1,102 @@
...
+-   X Axis labels
+-   Y Axis labels
+-   Legend
+
+Construct exposes following functions to handle the graph or its contents:
+
+-   `generate` - Called internally when instantiating the graph
+-   `loadContent` - Load a dataset in the graph
+-   `unloadContent` - Remove a dataset from the graph
+-   `resize` - Resizes the graph based on window resize event
```
swatikode: Are `generate` and  `resize` methods exposed to consumers or solely an internal methods. If so, I would indicate that somewhere

docs/core/Lifecycle.md
```diff
@@ -0,0 +1,102 @@
...
+For instance:
+
+-   A `Gantt` chart cannot be combined with a `Line` graph -> `Line` needs a X and Y axis whereas `Gantt` chart Construct only has X Axis when drawn.
+-   A `Timeline` graph cannot be combined with a `Line` graph -> `Line` needs a X and Y axis whereas `Timeline` graph doesn't have a Y Axis.
+-   `Line` graph can be combined with a `Bar` or `Paired Result`.
+
+## GraphContent
+
+`GraphContent` is the representation of the data-set drawn within the `Construct`. For instance, for `Line` graph, the `Construct` is the Axes, Labels and Legend while the `GraphContent` is the line itself.
+Separation of container and content allows us to:
```
swatikode: nit: I would put a white space on line 96 and place the `Separation of container....` text together with  `-Render multiple ...` text

PR #89, opened 2019-09-24
Fix: To not trigger popup for not truncated trackLabel (fixes #20)

### LINKED ISSUE #20: onClick triggers popup for Gantt Label items which are not truncated
**Describe the bug**
onClick triggers popup for Gantt Label items which are not truncated.

**To Reproduce**

1. Go to https://engineering.cerner.com/carbon-graphs/#/gantt/bar-types
2. Click on 'track label' for any gantt chart.
3. See error

**Expected behavior**
onClick should not trigger popup for Label which are not truncated. 
onClick should be triggered only for the truncated Label items. 

**Screenshots**
pop up for non truncated track label

pranav300:

-   [ ] Feature
-   [x] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x ] No

Updated Gantt examples to have `onclick` only for `trackLabel` which are long enough to get truncated

**Add any screenshots**
Screenshot 2019-09-24 at 3 48 37 PM


abhijit945:
There seems to be a misunderstanding here. 
The issue #20 is supposed to be for consumers, not the site per se. 
Currently track label click action is triggered regardless of whether they are truncated or otherwise. We would only want the `onClick` to be triggered when the label is truncated.

src/main/js/controls/Gantt/Track.js
```diff
@@ -98,9 +99,14 @@ const addTrackLabelEventHandler = (canvasSVG, labelObj) =>
...
                 );
             }
-            d3.select(this).on("click", () => {
-                labelObj.onClick(labelObj.display, d3.select(this));
-            });
+            if (
+                labelObj.display.length >
+                constants.DEFAULT_LABEL_CHARACTER_LIMIT
+            ) {
+                d3.select(this).on("click", () => {
```
abhijit945: The cursor pointer is [shown](https://github.com/cerner/carbon-graphs/blob/master/src/main/less/controls/Gantt.less#L6) only when the consumer provides `onClick` property via the `aria-disabled` attribute. So if you allow the code to pass through Ln 99 then it would show the cursor pointer but when you click nothing would show up. 

Ideally, it should be:

Truncated: 
image

Non-truncated: 
image

src/test/unit/controls/Gantt/Track-spec.js
```diff
@@ -166,10 +166,7 @@ describe("Gantt -> Track", () => {
                     expect(labelPath.getAttribute("aria-disabled")).toBe(
                         "false"
                     );
-                    expect(onClickPrimaryFunctionSpy.calledOnce).toBeTruthy();
-                    expect(onClickPrimaryFunctionSpy.getCall(0).args[0]).toBe(
-                        smallLabel
-                    );
+                    expect(onClickPrimaryFunctionSpy.calledOnce).toBeFalsy();
```
abhijit945: You can check for `aria-disabled` to be `true` here as well.

src/main/js/controls/Gantt/Track.js
```diff
@@ -98,9 +99,14 @@ const addTrackLabelEventHandler = (canvasSVG, labelObj) =>
                     !utils.isFunction(labelObj.onClick)
                 );
             }
-            d3.select(this).on("click", () => {
-                labelObj.onClick(labelObj.display, d3.select(this));
-            });
+            if (
+                labelObj.display.length >
```
abhijit945: There is a nifty helper for this `shouldTruncateLabel`

src/main/js/controls/Gantt/Track.js
pranav300: Fixed here https://github.com/cerner/carbon-graphs/pull/89/commits/869cb84d5e749a87f211ddc7135beec7d007283c.

src/test/unit/controls/Gantt/Track-spec.js
pranav300: Fixed here https://github.com/cerner/carbon-graphs/pull/89/commits/869cb84d5e749a87f211ddc7135beec7d007283c.

src/main/js/controls/Gantt/Track.js
pranav300: Fixed here https://github.com/cerner/carbon-graphs/pull/89/commits/869cb84d5e749a87f211ddc7135beec7d007283c.

abhijit945:
The issue #20 is titled for Bar Graph's Axis Info Row but the description is for Gantt by mistake. Ill create a separate issue for Bar Graph.

abhijit945:
Created https://github.com/cerner/carbon-graphs/issues/90 for Bar Graph Axis Info Row

PR #98, opened 2019-10-04
Fix: prevents onclick trigger for non truncated axis row label (fixes #90)

### LINKED ISSUE #90: onClick triggers popup for Axis Label Row labels which are not truncated
**Describe the bug**
onClick triggers popup for Axis Label Row items which are not truncated

**To Reproduce**

1. Go to https://engineering.cerner.com/carbon-graphs/#/bar/axis-info-text-labels
2. Click on 
image

3. Now click on 
image

4. Popup should not load

**Expected behavior**
onClick should be triggered only for truncated labels. 

**Screenshots**
image

-   OS: [e.g. iOS] Windows 10
-   Browser [e.g. chrome, safari] Chrome

**Additional context**
N/A

avinashg1994:

-   [ ] Feature
-   [X] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

Added a condition which allows attaching of an `onClick` event and setting of `aria-disabled` only when the `axisInfoRow` label is truncated.

**Add any screenshots**
When clicked on truncated label
Screenshot 2019-10-04 at 6 39 03 PM

When clicked on non-truncated label
Screenshot 2019-10-04 at 6 39 47 PM


src/main/js/controls/Bar/helpers/axisInfoRowHelpers.js
```diff
@@ -58,17 +58,28 @@ const createAxisInfoRowLabel = (
...
-            );
-        });
+        .attr("aria-hidden", config.shownTargets.indexOf(uniqueKey) < 0);
+    if (
+        shouldTruncateLabel(
+            textLabelList.value.label.display,
+            textLabelList.value.characterCount
+        )
+    ) {
+        axisInfoRow
```
abhijit945: +1 for Primary `display` but we need the same behavior with `secondaryDisplay` as well. 

Try adding a large text for **only** `secondaryDisplay` and see what happens with the above code enabled. Also check out `label` row here: https://github.com/cerner/carbon-graphs/blob/master/docs/controls/Bar.md#axis-info-row-value on how to update the values. I think there is an example in the site which you can play around with!

avinashg1994: I agree, in the scenario when the value of only `secondaryDisplay` (not `display`) exceeds the characterCount, the onClick is disabled.
So, implemented the same behaviour for `secondaryDisplay` as well.

abhijit945:
@avinashg1994 
Can you include some of your teammates for this review? Need one more +1 for this.

PR #108, opened 2019-10-28
Fix: Show region when only one paired result is displayed (fixes #3)

### LINKED ISSUE #3: Paired Result - Multiple Regions

benbcai:

-   [ ] Feature
-   [X] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [X] No

Fixes #3 
The regions of a paired result graph were not displayed when only one pair result is selected. This was because a pair result would also contain more than one key and the `isSingleTargetDisplayed` would return true only if there is only one key. A method called `isSinglePairedResultTargetDisplayed` is created to check if only one pair result is displayed. 

**Add any screenshots**
Screen Shot 2019-10-28 at 12 43 13 PM


abhijit945:
Looks good to me 👍

PR #107, opened 2019-10-23
Update: Constant padding bottom (fixes #81)

pranav300:
Fixes #81

**What is the purpose of this pull request? (Check any that's applicable)**

-   [ ] Feature
-   [x] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes
-   [ ] Changes an existing functionality
-   [ ] Other, please explain:

**Does this introduce a breaking change?**

-   [ ] Yes
-   [x] No

**What changes did you make? (Give an overview)**
Updated constant padding-bottom to be 2.5
**Add any screenshots**
OS: Windows
Browser: Chrome

Before:
chrome before

After:
chrome after

Browser: Firefox

Before: 
firefox before

After:
firefox after

OS: Mac
Browser: Chrome

Before:
mac chrome before

After:
mac chrome after

Issue #81, opened 2019-09-18
Timeline X Axis label padding has changed


abhijit945:
**Describe the bug**
X Axis label has moved down and getting cut-off

**To Reproduce**

1. Go to https://engineering.cerner.com/carbon-graphs/#/timeline/simple

**Expected behavior**

Label should be clearly visible

**Screenshots**

![image](https://user-images.githubusercontent.com/4908886/65151187-2dc5f200-d9eb-11e9-863a-f7f0ff63e325.png)

**Desktop (please complete the following information):**

-   OS: [e.g. iOS] Windows 10
-   Browser [e.g. chrome, safari] Chrome

pranav300:
Investigation
What I have found is that this issue is `OS` specific. In windows chrome and IE the `config.LabelHeights.x` has a value 19, whereas in Mac chrome the value is 17, And for firefox, the difference increases to `3.5`. I have found two solutions for it

1. Increase constant padding `top` from `10` to `20`. But there will always be difference in the value returned by the function `getAxisLabelHeight`

2. To keep a check for OS i.e Win32 and within this have one more check for firefox.
```diff
const getAxisLabelHeight = (label) => {
    const dummy = d3.select("body").append("div");
    const svg = dummy.append("svg");
    const grouper = svg.append("g");
    buildAxisLabel(grouper, label);
-    const height = grouper.node().getBoundingClientRect().height;
+    let height = 0;
+    if (window.navigator.platform === "Win32") {
+        if (window.navigator.userAgent.indexOf("Firefox") > -1) {
+            height = grouper.node().getBoundingClientRect().height - 3.5;
+        } else {
+            height = grouper.node().getBoundingClientRect().height - 2
+        }
+    } else {
+        height = grouper.node().getBoundingClientRect().height;
+   }
    dummy.remove();
    return height;
};
```

**Mentions**
@abhijit945

abhijit945:
As mentioned offline:
I am not sure if we would want to go the browser specific route. Can you try using getBBox instead of getBoundingClientRect and see if we can get away from this browser specific behavior? 
https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement/getBBox
https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

pranav300:
I tried using `getBBox` but, got the same results. So, planning to go ahead with the browser-specific route.

ShettyAkarsh:
[...]

PR #125, opened 2019-12-12
Fix: Y2 axis going out of scope when hasCustomPadding is true.

### LINKED ISSUE #110: Y2 label getting truncated when hasCustomPadding is true.
**Describe the bug**
 Y2-axis label is getting truncated when `hasCustomPadding`
 is true.

**To Reproduce**

1. Go to the code and add custom padding to [this](https://github.com/cerner/carbon-graphs/blob/master/dev/examples/controls/line.js#L34) similar to that in constants.
3. run the local server
4. Go to this link http://localhost:9991/#/line/y2-axis

**Expected behavior**
Y2-axis label should be visible with hasCustomProperty.

**Screenshots**
Screenshot 2019-11-12 at 3 27 31 PM

-   OS: IOS
-   Browser: Chrome

pranav300:

-   [ ] Feature
-   [x] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

Fixes #110 #81

While working on issue #110, I found that PR #107 introduced a bug in the graph construct. As there was slight clipping of the y-axis and y2-axis at the top. Due to the change in padding-bottom in constants. So, I am fixing that as well in this PR.

**Add any screenshots**
Y2 graph in DevSite.
image


src/main/js/controls/Timeline/helpers/creationHelpers.js
```diff
@@ -91,7 +91,7 @@ const getXAxisXPosition = (config) => config.padding.left;
  * @returns {number} Position for the axis
  */
 const getXAxisYPosition = (config) =>
-    (config.padding.top + config.padding.bottom) * 2;
+    config.padding.top * 1.5 + config.padding.bottom;
```
abhijit945: Why are we doing this? Didnt we fix a bug for firefox recently that increased the padding top? 
My concern is that, the unit tests no longer give us the protection in this case. 
Unit tests have a number 300. Fix 1 increases top to 20 -> Unit test now checks for 320.
Fix 2 decreases top by half -> Unit test now checks for 160. i.e. this no longer tests the Fix 1

pranav300: > Why are we doing this? Didn't we fix a bug for firefox recently that increased the padding-top?

IN PR #107, we had decreased the padding-bottom from `5` to `2.5`. Which made it work for timeline but as a result, all the graphs created from graph-construct had their top Y and Y2 axis ticks getting clipped. So, I have reverted the padding-bottom back to `5` and to solve the timeline issue this change `config.padding.top * 1.5 + config.padding.bottom;` was made. As this would only affect the timeline. 
What do you think? @abhijit945

abhijit945: I am good with that as long as we are only changing the unit tests for impacted functionality.

abhijit945:
Please update the commit message to a single commit with appropriate commit message.

abhijit945:
@pranav300 This has my +1, you can get another reviewer to get review it and potentially merge it 👍

src/main/js/helpers/constants.js
```diff
@@ -149,7 +149,7 @@ export const AXES_ORIENTATION = {
 export default {
     PADDING: {
         top: 10,
-        bottom: 2.5,
+        bottom: 5,
```
abhijit945: Does doing this fix the issue happening in the site right now (_The topmost values in the Y Axis are cut-off_)? 

image

pranav300: Yes, it fixes it.

PR #126, opened 2019-12-12
Fix: Padding Top on Timeline changes the width (fixes #117)

### LINKED ISSUE #117: Padding Top on Timeline changes the width of the timeline
**Describe the bug**
When you add padding top value above 0 and all the other padding properties are 0. It seems to be changing the width of the timeline, thus causing our other components out of sync with it.

**To Reproduce**
see running example here:
https://codesandbox.io/s/nervous-hofstadter-g07l2

**Expected behavior**
Only the top padding should be applied, the width should remain the same.

**Screenshots**
Screen Shot 2019-11-21 at 17 33 34 1

-   OS: Mac
-   Browser: Chrome

saket2403:

-   [ ] Feature
-   [x] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

Removed dependency of timeline's width on padding top and bottom. Fixes #117.

**Add any screenshots**
Screenshot 2019-12-12 at 6 10 15 PM


abhijit945:
@saket2403 Can you go over which issue this fixes and what was the issue/root cause?

saket2403:
@abhijit945 . The root cause of the issue was that `getXaxisYPosition` was used in the calculation of X-axis width in the function `getXAxisWidth`. I think The width of the X axis is independent of top and bottom padding of X axis. 
I am seeing something similar in Gantt Chart also [here](https://github.com/cerner/carbon-graphs/blob/7d3d238eff2727fa7316f57d70c285624442d1a5/src/main/js/controls/Gantt/helpers/creationHelpers.js#L198-L211). This will also need a similar fix. Should I fix this in this PR or log a separate issue for gantt chart and fix it.

abhijit945:
Let's do that in another PR and focus this for the Timeline alone.

abhijit945:
Can you squash all the commits to a single commit? 
Make the commit text to `Fix: Padding Top on Timeline changes the width (fixes #117)`

abhijit945:
@saket2403 Can you resolve the conflict, before I merge this?

PR #131, opened 2019-12-20
Fix: Opacity for Hashed Activity Bars for Gantt (fixes #113)

### LINKED ISSUE #113: Change background hash bar opacity percentage for accessibility contrast
Background hash bar at 30% opacity does not pass 3:1 contrast even with straight black. We need to update that percentage to allow for a darker gray, and/or change the color and the opacity as the color is locked down at the moment.

### LINKED ISSUE #123: Change non-hash background bars to include a border for accessibility
The current background bar design does not allow for high-enough contrast for background bars against the background of the gantt. The library goal of a background bar is to have an indication of a status (like total duration, expected duration, life of the order, etc.) but not be overpowering to the foreground bars or the primary information of duration (like effort length, dialysis duration, or medication administration duration). In order to create a visible foreground and background bar and events/actions on foreground bars, the background bar fill must remain relatively light, but I would suggest adding a border to it to pass contrast against the row background. This will allow for the foreground bar to pass against the background bar and the background bar border to pass against the row background. 

Example: 
image

abhijit945:

-   [ ] Feature
-   [ ] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes
-   [X] Changes an existing functionality

-   [ ] Yes
-   [X] No

`Fix: Opacity for Hashed Activity Bars for Gantt (fixes #113)`

**Before**

image

**After**

image

`Fix: Border Background bars Gantt for a11y (fixes #123)`

**Before**

image

**After**

image


laurenbock:
added 1px border on background which is full opacity so consumers can pass that border against white background for contrast. added 50% opacity on hash background bar so consumers can pass it black to get a high enough contrast against white background.

PR #140, opened 2020-01-07
Fix: Graph API now preserves order of content loaded (fixes #13)

pranav300:
**What is the purpose of this pull request? (Check any that's applicable)**

-   [ ] Feature
-   [X] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes
-   [ ] Changes an existing functionality
-   [ ] Other, please explain:

**Does this introduce a breaking change?**

-   [ ] Yes
-   [X] No

**What changes did you make? (Give an overview)**
Update `clickHandler` for the legend of graphs (Line, PairedResult and Scatter) created through Graph API. So, as to stop the removal of DOM elements when the legend button is toggled.
Note: No updates were made in Bar graph as I didn't observe the issue #13  in it.

**Add any screenshots**
Before:
Before

After:
after

PR #164, opened 2020-02-12
Fix: Fixed graphs to not render data points where y value is null.

### LINKED ISSUE #161: Toggling of Non-Contiguous Line Legend Adds a Data Point
When the legend item is toggled off and then on again for a non-contiguous line, a random point will appear at the bottom of the graph. 

When the legend item is toggled back on, no new points should be added to the graph 

Screen Shot 2020-02-04 at 11 44 34 AM

pranav300:

-   [ ] Feature
-   [X] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [X] No

I have put conditions in `Gantt, Timeline, Line, Scatter` so that dataPoints are not rendered when dataPoint y value is null and removed function `shouldHideDataPoints`, as it seemed unrequired. Because on the first load all the data is shown. And the hiding and visibility of the dataPoints are later taken care of by clickHandler update in [this PR](https://github.com/cerner/carbon-graphs/pull/140).

closes #161


src/main/js/controls/Gantt/helpers/actionHelpers.js
```diff
@@ -276,8 +262,24 @@ const drawActionDataPoints = (scale, config, canvasSVG) =>
         .classed(styles.pointGroup, true)
         .each(function(dataPoint, index) {
             const dataPointSVG = d3.select(this);
-            renderSelectionPath(scale, config, dataPointSVG, dataPoint, index);
-            renderDataPointPath(scale, config, dataPointSVG, dataPoint, index);
+            if (dataPoint.y !== null) {
```
abhijit945: There is nothing wrong with this approach but we could make use of `d3.select.filter` function instead?
It would work this way:
```js
d3.select("body")
    .selectAll("p")
    .data(
        [4, 8, 15, 16, 23, 42].filter(function(d) {
            return d < 10;
        })
    )
    .enter()
    .append("div")
    .text(function(d) {
        return d;
    });
```

pranav300: done here https://github.com/cerner/carbon-graphs/pull/164/commits/b4d0cd4a3fe85882be8c7acfd2c8cde97ca5dda0.

src/test/unit/controls/Scatter/ScatterLoad-spec.js
```diff
@@ -511,10 +509,10 @@ describe("Scatter - Load", () => {
                     );
                     expect(scatterGroup.length).toBe(1);
                     expect(
-                        scatterGraphContainer
-                            .querySelectorAll(`.${styles.point}`)[2]
-                            .getAttribute("aria-hidden")
-                    ).toBe("true");
+                        scatterGraphContainer.querySelectorAll(
+                            `.${styles.point}`
```
abhijit945: Why is this impacting current tests?

src/main/js/controls/Scatter/helpers/helpers.js
```diff
@@ -182,7 +182,7 @@ const draw = (scale, config, canvasSVG, dataTarget) => {
     const pointPath = scatterSVG
         .select(`.${styles.currentPointsGroup}`)
         .selectAll(`.${styles.point}`)
-        .data(getDataPointValues);
+        .data(getDataPointValues(dataTarget).filter((d) => d.y != null));
```
abhijit945: `!==`

src/test/unit/controls/Scatter/ScatterLoad-spec.js
pranav300: The current tests were checking whether a point exists in the DOM whose y value is null and having `aria-hidden` to be `true`. But after the changes that point is not present in the DOM itself, So the tests were failing. 
In the tests now, it is checking the total no. of `carbon-data-points` present in the DOM. Which is then compared with the length of dataArray after subtracting the no of data points with y value null.

src/main/js/controls/Scatter/helpers/helpers.js
pranav300: done here https://github.com/cerner/carbon-graphs/pull/164/commits/2ea11e54b41f833eab9355a6016affc125191f3d.

pranav300: done here https://github.com/cerner/carbon-graphs/pull/164/commits/2ea11e54b41f833eab9355a6016affc125191f3d.

abhijit945:
Dont forget to add `fixes (#161)` to your PR, that way GitHub closes the issue when this PR is closed.

PR #165, opened 2020-02-13
Fix: Fixed timeline X-Axis width to show end data (fixes #155)

### LINKED ISSUE #155: Timeline does not render data points(values) which are the same as the upper limit
**Describe the bug**
Data points that are exactly the same as the upper limit do not display on the timeline. Whereas data points exactly the same as the lower limit do display.

**To Reproduce**
See working example of issue here:
https://codesandbox.io/s/data-points-same-as-end-date-dont-display-wnc99

**Expected behavior**
The data point should display half of the icon since its on the edge of the timeline.

**Screenshots**
Screen Shot 2020-01-25 at 11 12 48

- OS: Mac
- Browser: Chrome

saket2403:

-   [ ] Feature
-   [x] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

Added half of width of first and last ticks as left and right padding if xAxis width is more than canvas width.

**Before Fix**
image

**After Fix**
image


src/main/js/controls/Timeline/Timeline.js
```diff
@@ -197,6 +197,20 @@ class Timeline extends Construct {
             );
         createDefs(this.config, this.svg);
         createAxes(this.axis, this.scale, this.config, this.svg);
+        const ticks = document.getElementsByClassName("tick");
+        const firstTick = ticks[0].getBoundingClientRect().width;
+        const lastTick = ticks[ticks.length - 1].getBoundingClientRect().width;
+        const xAxis = document
+            .getElementsByClassName("carbon-axis carbon-axis-x")[0]
+            .getBoundingClientRect().width;
+        if (xAxis + firstTick / 2 + lastTick / 2 > this.config.canvasWidth) {
```
abhijit945: In the description  you say `Added half of width of first and last ticks as left and right padding if xAxis width is more than canvas width.` 
Why is this necessary?
Why is X Axis width more than canvas width?

saket2403: Actually the sum of X-Axis width and half of first and last ticks becomes more than canvas width.

src/main/js/controls/Timeline/Timeline.js
```diff
@@ -197,6 +197,20 @@ class Timeline extends Construct {
...
+            d3RemoveElement(this.graphContainer, "defs");
```
abhijit945: Why delete defs when you are creating it in line 198?

src/main/js/controls/Timeline/Timeline.js
abhijit945: Right, you have a fix here but I am more interested in knowing why the issue is created in the first place so that we can better understand it.

saket2403: Let's say the canvas and X axis both span from 0 to 1200px so the end ticks which start at points 0 and 1200 get cut off because they they have their own widths , so we need some padding values to incorporate these end ticks.

src/main/js/controls/Timeline/Timeline.js
saket2403: I am deleting both the X axis and defs and then re creating them with the new padding values, because deleting only the X axis gives console errors.

src/main/js/controls/Timeline/Timeline.js
abhijit945: So based on what you are saying, if the end ticks werent there then would this be fixed?

swatikode:
The snapshot in the description shows both the 01:00 and 16:00 ticks correctly. Why is it indicated that the 16:00 tick is missing? I can't seem to understand if the snapshot is for the fix or the issue. Can you clarify this? Also, to better understand what is causing the issue in the first place can you post a snapshot of the ticks being cut off from both the left and right ticks?

saket2403:
> The snapshot in the description shows both the 01:00 and 16:00 ticks correctly. Why is it indicated that the 16:00 tick is missing? I can't seem to understand if the snapshot is for the fix or the issue. Can you clarify this? Also, to better understand what is causing the issue in the first place can you post a snapshot of the ticks being cut off from both the left and right ticks?

Added screenshots of before and after the fix, hope this helps.

src/test/unit/controls/Timeline/Timeline-spec.js
```diff
@@ -235,14 +235,18 @@ describe("Timeline", () => {
             input.content = null;
             data.tasks = [];
             const cfg = timeline.config;
+            const ticks = document.getElementsByClassName("tick");
+            const leftValue = ticks[0].getBoundingClientRect().width / 2;
+            const rightValue =
+                ticks[ticks.length - 1].getBoundingClientRect().width / 2;
```
swatikode: Why are the expected values calculated here? You are using the same formula as used in the source code that is going to work. This is not how you should be testing.

swatikode:
The width of the canvas is calculated in the method [setCanvasWidth](https://github.com/cerner/carbon-graphs/blob/ce28fd9b69e32beaa6fd7206c3deecb3919f2553/src/main/js/controls/Graph/Graph.js#L57) and it takes into the account the width of the container, the container the consumer chooses to place the graph and so the width could be anything. What made you decide that adding a padding 1/2 of the left and right tick will be enough at all times? Try choosing a container with different widths and confirm the code change, also add the corresponding unit tests.

src/test/unit/controls/Timeline/Timeline-spec.js
saket2403: Fixed 675cd0b

saket2403:
> The width of the canvas is calculated in the method [setCanvasWidth](https://github.com/cerner/carbon-graphs/blob/ce28fd9b69e32beaa6fd7206c3deecb3919f2553/src/main/js/controls/Graph/Graph.js#L57) and it takes into the account the width of the container, the container the consumer chooses to place the graph and so the width could be anything. What made you decide that adding a padding 1/2 of the left and right tick will be enough at all times? 

I had tested different scenarios with and without container before raising the PR all seems to work fine. As for the solution proposed I figured out that a minimum amount of padding is required to prevent the ticks from getting truncated, whose value I calculated it to be equal to half the widths of 1st and last padding in different scenarios so added a fix accordingly.

src/test/unit/controls/Timeline/Timeline-spec.js
saket2403: Fixed here e607e6e

ShettyAkarsh:
@saketh, so currently you are taking the ticks value from 1:00 AM to 4:00 PM. so what happens if the number of ticks are still increased ? let say to 6:00 PM ? so in this case still the canvas width will be more than the x-axis width. so will your logic of taking half the padding from left and right and then adding to x-axis width still works ?

ShettyAkarsh:
Not sure if this works here. so have you tried something similar to this logic. in the case if x-axis width is greater than canvas width ? 

if (x-axis width > canvas width)
{
    Diff width = (x-axis width - canvas width)
    New canvas width =  canvas width + Diff width
}

other wise 

if (x-axis width > canvas width)
{
    Diff width = (x-axis width - canvas width)
    New canvas width =  Diff width / 2
    // Assign the New canvas width as right and left padding values for canvas.
}

saket2403:
> Not sure if this works here. so have you tried something similar to this logic. in the case if x-axis width is greater than canvas width ?
> 
> if (x-axis width > canvas width)
> {
> Diff width = (x-axis width - canvas width)
> New canvas width = canvas width + Diff width
> }
> 
> other wise
> 
> if (x-axis width > canvas width)
> {
> Diff width = (x-axis width - canvas width)
> New canvas width = Diff width / 2
> // Assign the New canvas width as right and left padding values for canvas.
> }

Changing canvas Width is not the right approach as it depends on the size of the container chosen by the user so increasing canvas Width can make its width more than container, thats why re sizing X-axis to fit inside the canvas seems better.

saket2403:
> @saket2403 ,so currently you are taking the ticks value from 1:00 AM to 4:00 PM. so what happens if the number of ticks are still increased ? let say to 6:00 PM ? so in this case still the x-axis width will be more than the canvas width. so will your logic still works ?

I am taking widths of the first and last ticks irrespective of their values, only condition being [this](https://github.com/cerner/carbon-graphs/pull/165/files#diff-db00fd9263d1463026a5bfc662dd1526R206). So it will work for any time value.

src/main/js/controls/Timeline/Timeline.js
saket2403: Yes in certain cases the X axis shortens to an extent that end ticks are not displayed and issue is fixed.

ShettyAkarsh: `xAxis + firstTick / 2 + lastTick / 2` add it inside the braces to make it more clear.

PR #190, opened 2020-03-24
New: Honor Height Regardless of Outlier Check (fixes #74)

### LINKED ISSUE #74: Outliers/Viewable area
There are cases in which including all data points face up will impact the usability/interpretation of the lines. Outliers could impact the slope of the line causing it to be very flat and unusable. This request is to show a viewable area based on fixed ranges and if data points appear outside of that indicate to the user that there are points outside the viewable area via running the line off the edge of the viewable area paired with a banner notifying the user of the data outside of the viewable graph. 

**Reasoning:** The reason we wanted to avoid showing all data points is change blindness. If the user had a standard graph view they used for every patient, but the y-axes were slightly different from patient to patient, they might not see the change in axis values and could rely on the slope of the line instead. If points appear off the edge of the graph, drawn lines themselves are indication enough since they go out of the current view. Allowing the user to access what those points are is necessary for them either in progressive disclosure. 

image

SharathAlur:
https://github.com/cerner/carbon-graphs/issues/74

Ability to avoid padding and auto-calibration of the y-axis.

-   [x] Feature
-   [ ] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

Added a property **allowCalibration** to the graph. When this is set to false, avoid y-axis range update and set clamp to false.

**Add any screenshots**
Disable_Calibration


docs/core/Graph.md
```diff
@@ -31,21 +31,22 @@ Graph is a set of graphs that share a common set of base visualization structure
...
+| showVGrid        | boolean  | true                                          | Toggle to show vertical grid                                                         |
+| showHGrid        | boolean  | true                                          | Toggle to show horizontal grid                                                       |
+| dateline         | array    | []                                            | Refer [Dateline](../helpers/Dateline.md)                                             |
+| pan              | object   | {}                                            | Refer [Panning](../controls/Panning.md)                                              |
+| padding          | object   | `{ top: 10, bottom: 5, left: 30, right: 50 }` | Refer [Padding](../controls/Padding.md)                                              |
+| showNoDataText   | boolean  | true                                          | Toggle to show no data text                                                          |
+| allowCalibration | boolean  | true                                          | Toggle to allow calibration to adjust the y axis                                     |
```
ryanthemanuel: Do we want to call this disableCalibration and have the default be false?  I think I like this better and it is maybe a little clearer, but I'm open to thoughts.

SharathAlur: If we analyze all the props are having positive words, even though they are optional. I feel it is better to follow the same trend. Let me know your thoughts.

mraplumb: I would agree with trying to equate a truthy value with "on" and a falsey value with "off"

mraplumb:
@SharathAlur please resolve the conflicts introduced by the merge of another PR

SharathAlur:
@mraplumb I have resolved the conflict and also updated the data to correctly show/hide the shapes per data type in the dev site.

PR #248, opened 2020-06-24
New: Suppress Tick Values Trailing Zeros (fixes #234)

### LINKED ISSUE #234: Allow consumers to use a whole number instead of decimal
When a Y-axis range contains decimals, carbon always displays all numbers in that range with a decimal. 
MicrosoftTeams-image (33)

Give consumers the ability to display whole numbers and decimals in a range. This is to help users distinguish between a change to a new whole number within that range. 
image

JananiGunasekaran:

-   [x] Feature
-   [ ] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

Introduced a new prop `suppressTrailingZeros` inside the axis object, thus enabling consumer to handle suppressing the trailing zeros separately for each axis (X, Y and Y2). When this prop is set to true for any axis, it will suppress the axis's tick values's trailing zeros when default d3 tick formatting is used.
If the consumer specifies the tick values, they must specify '~' just before the format type. `Eg: format: "~r"` in order to suppress tick values's trailing zeros ( Updated the Ticks documentation ).

**Add any screenshots**
Default behaviour / When `suppressTrailingZeros` is set to false for each axis - (X, Y and Y2) :

image

When `suppressTrailingZeros` is set to true for each axis - (X, Y and Y2) :

image


src/main/js/helpers/axis.js
```diff
@@ -175,17 +175,33 @@ const createAxisReferenceLine = (axis, scale, config, canvasSVG) => {
  * @returns {object} - Scaled axes object
  */
 const getAxesScale = (axis, scale, config) => {
+    let tickFormatToTrimTrailingZeros;
+
+    if (
+        config.suppressTrailingZeros &&
+        config.axis.x.type === AXIS_TYPE.DEFAULT &&
+        utils.isUndefined(config.axis.x.ticks.format)
```
Veera81, line 183: Can you add a comment on why we are doing this way?

You have added below scenario here for x-axis
- If the consumer

    1. `suppressTrailingZeros` prop is provided
    2. when axis type is default
    3. x-axis ticks format is undefined

Are we going to cover the below scenarios as well here?

What would happen if 

- If the consumer 

    1. `suppressTrailingZeros` prop is provided
    2. when axis type is default
    3. x-axis ticks format is **defined**

What would happen if 

- If the consumer

    1. `suppressTrailingZeros` prop is **not** provided
    2. when axis type is default
    3. x-axis ticks format is **defined**

src/main/js/helpers/axis.js
```diff
@@ -205,26 +221,67 @@ const getAxesScale = (axis, scale, config) => {
             config.axis.y2.domain.upperLimit - config.axis.y2.domain.lowerLimit
         );
 
+    if (
+        config.suppressTrailingZeros &&
```
Veera81, line 225: Why did not require a Y-axis type default check here like we added for the x-axis?

src/main/js/helpers/axis.js
```diff
@@ -205,26 +221,67 @@ const getAxesScale = (axis, scale, config) => {
...
+ * @private
+ * @param {object} axis - Scaled axes object
+ * @returns {string} tick format string
+ */
+const tickFormatter = (axis) => {
+    const defaultTickFormat = axis.scale().tickFormat().toString();
+    return `${defaultTickFormat.slice(
+        0,
+        defaultTickFormat.length - 1
+    )}~${defaultTickFormat.slice(defaultTickFormat.length - 1)}`;
```
Veera81, line 282: Can you add a comment here what exactly we are returning? and provide me one example of return value for this method?

src/test/unit/controls/Graph/GraphAxes-spec.js
```diff
@@ -408,6 +408,47 @@ describe("Graph - Axes", () => {
             const rangeY2 = 250;
             expect(getAverageTicksCount(rangeY, rangeY2)).toEqual(5);
         });
+        it("Suppresses tick values's trailing zeros when default d3 tick formatting is used and suppressTrailingZeros is set to true", () => {
```
Veera81, line 411: Can we add below testing considerations as well if not converted in the existing test cases? and also Are these cover X and Y and Y-2 axis well?  

- when default d3 tick formatting is **not used** `suppressTrailingZeros` is set to **true**.

- when default d3 tick formatting is **used** suppressTrailingZeros is set to **false**

- when default d3 tick formatting is **not used** `suppressTrailingZeros` is set to **false**

Probably, we may add a test case if x-axis type is default and not default based on the changes added

```js
            1. describe("when default d3 tick formatting is not used suppressTrailingZeros is set to true", () => {
                it("should suppresses tick values's trailing zeros ", () => {
                ...................................................................................

            2. describe("when default d3 tick formatting is used suppressTrailingZeros is set to false", () => {
                it("should not suppresses tick values's trailing zeros ", () => {
                ...................................................................................

            3. describe("when default d3 tick formatting is not used suppressTrailingZeros is set to false", () => {
                it("should not suppresses tick values's trailing zeros ", () => {
                ...................................................................................
```

src/main/js/helpers/axis.js
```diff
@@ -205,26 +221,67 @@ const getAxesScale = (axis, scale, config) => {
...
+                ? config.axis.y.ticks.format
+                : tickFormatToTrimTrailingZeros
+        ),
         config.ticksCount
     );
 
     if (hasY2Axis(config.axis)) {
+        if (
+            config.suppressTrailingZeros &&
+            utils.isUndefined(config.axis.y2.ticks.format)
```
Veera81, line 248: Can you add a comment here as well?

docs/helpers/Ticks.md
```diff
@@ -23,10 +23,10 @@ If there is only 1 axis in the graph, the tick intervals are calculated automati
...
+| Property Name | Expected | Description                                                                                                                                                                                                                                              |
+| ------------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| values        | array    | Ticks values in the axis. Each tick will have an associated grid line. If they are not provided, optional tick values need to exist.                                                                                                                     |
+| format        | string   | [Formatter](https://github.com/d3/d3-time-format/blob/master/README.md#locales) for tick values. `Note:` This is mandatory if values are provided. To suppress tick values's trailing zeros, specify '~' just before the format type. `Eg: format: "~r"` |
```
Veera81, line 29: Why do we only this specific `~` operator?

JananiGunasekaran: According to d3-format documentation [here](https://github.com/d3/d3-format/blob/master/README.md#locale_format), they have specified that '~' character will trim the insignificant trailing zeros.
The below is what's defined in the documentation:
"""" The `~` option trims insignificant trailing zeros across all format types. This is most commonly used in conjunction with types r, e, s and %. For example:

d3.format("s")(1500);  // "1.50000k"
d3.format("~s")(1500); // "1.5k"            """"""

src/main/js/helpers/axis.js
JananiGunasekaran: Added comment [here](https://github.com/JananiGunasekaran/carbon-graphs/commit/655b2444b4ec1bee262ef03c4e59f429f3f9b6cd). 

According to d3-format documentation [here](https://github.com/d3/d3-format/blob/master/README.md#locale_format), the ~ character will trim the insignificant trailing zeros.
So we get the default d3 tick format and insert '~' just before the format type.

I have console logged the default d3 tick format and the value returned by this tickFormatter () method here:
image

src/main/js/helpers/axis.js
JananiGunasekaran: Added comment [here](https://github.com/JananiGunasekaran/carbon-graphs/commit/cd62efa8c9feea2ab4a5585cb63fa492169425ae).

src/main/js/helpers/axis.js
JananiGunasekaran: I have not added "axis-type == default" check for y-axis since there can be only number based axes values for y axis. The "axis-type == default" check is required for x-axis since there can be 2 types of axes values like time series or number based values for x-axis.

Added comment for the conditional statement [here](https://github.com/JananiGunasekaran/carbon-graphs/commit/7d888e61d33050d9b0099e5d57adac17c5cb1b17).

src/main/js/helpers/axis.js
JananiGunasekaran: Added comment [here](https://github.com/JananiGunasekaran/carbon-graphs/commit/4d775051c01587056c48fd5f9ad5d05699b247e3).

All the scenarios mentioned above are being covered in the code:

1. If the consumer 

- suppressTrailingZeros prop is **provided**
- when axis type is default
- x-axis ticks format is **defined**

------- The if condition will not get satisfied and it will execute the next statement after the loop [here](https://github.com/JananiGunasekaran/carbon-graphs/commit/19249fbfd155a12c377f2b9c1f59042777167551#diff-c116b52d2491f4d3b18768a42e2a7a23R191). This statement will use the x-axis tick format provided by the consumer to format the tick values of the axis. As mentioned in the [ticks.md](https://github.com/JananiGunasekaran/carbon-graphs/commit/19249fbfd155a12c377f2b9c1f59042777167551#diff-8a20b2ec7dd8fdd32c7802f2fcdee2b6R29) docs, the consumer must specify the '~' character if they want to suppress the trailing zeros when they specify the tick format.

2. If the consumer

- suppressTrailingZeros prop is **not provided**
- when axis type is default
- x-axis ticks format is **defined**

------- The if condition will not get satisfied and hence it will execute the next statement which will use the x-axis tick format provided by the consumer to format the tick values of the axis. If the consumer is going to specify the tick format, they have the option to decide whether they want to suppress the trailing zeros or not, by specifying the '~' character in their tick format.

3. There is also another scenario:
If the consumer

- suppressTrailingZeros prop is **not provided**
- when axis type is default
- x-axis ticks format is **undefined**

-------- The if condition will not get satisfied and hence it execute the next statement which will just use the default d3 tick format to format the tick values without suppressing the trailing zeros since both suppressTrailingZeros is not provided and x-axis ticks format is undefined.

src/test/unit/controls/Graph/GraphAxes-spec.js
JananiGunasekaran: I have added unit test cases for the scenarios [here](https://github.com/JananiGunasekaran/carbon-graphs/commit/1914cd49bbb91bc2b5b66782c74be5ba5475f5c1). Also a combination of `suppressTrailingZeros` (either set to true/false) & default d3 tick formatting is **not used** will not arise. Because `suppressTrailingZeros` is only to trim the trailing zeros when default d3 tick formatting is used.

I have added test cases for the following scenarios:
When default d3 tick formatting is used and suppressTrailingZeros is set to true
When default d3 tick formatting is used and suppressTrailingZeros is set to false
When default d3 tick formatting is not used and consumer specifies ~ in tick format
When default d3 tick formatting is not used and consumer does not specify ~ in tick format

I have added unit test cases for all the above 4 scenarios for x, y and y2-axes as well. I have covered the scenario when x-axis type is default. The other scenario which is x-axis type is TIME_SERIES need not be taken into account here since the values are going to be time based, so we need not worry about trailing zeros in this case.

docs/helpers/Ticks.md
Veera81: Thanks for the clarification and d3-format documentation

ji2307:
@JananiGunasekaran can you elaborate on the decision to support suppressing trailing zero's on the x-axis?  My initial thought was this would only apply to Y and Y2.

JananiGunasekaran:
@Veera81  The Travis CI - Pull Request was failing after my last commit. So i deleted the node modules and built the project again and ran "npm update". I committed the package.json and package-lock.json files in this [commit](https://github.com/JananiGunasekaran/carbon-graphs/commit/5796f456890be337bbd58e89edf3e17f3bc8781a). But i am not sure why the Travis CI - Pull Request is failing still.

Veera81:
@JananiGunasekaran Looks like there are 4 tests are failing in build. Can you run the tests in locally and see if any tests are failing

JananiGunasekaran:
> @JananiGunasekaran Looks like there are 4 tests are failing in build. Can you run the tests in locally and see if any tests are failing

@Veera81 I ran the tests after the npm update. All the tests are passing. I missed to mention this in my earlier comment. I ran the tests now as well, they are passing fine locally.
image

JananiGunasekaran:
@Veera81 I have made necessary changes to have the prop inside the axis object to handle suppressing the trailing zeros separately for each axis - (X, Y and Y2) in this [commit](https://github.com/JananiGunasekaran/carbon-graphs/commit/15d6b375ba40495f94c0d15c40109b2a87bec3e3).

PR #292, opened 2020-08-25
Update: Reduce gap in betweeen y-axis and y-axis label(fixes #30)

### LINKED ISSUE #30: Reduce spacing between y-axis and the y-label
**Is your feature request related to a problem? Please describe.**
We have implemented the carbon line graph in one of our applications and noticed that the y-label is a bit farther from y-axis than necessary. 

**Describe the solution you'd like**
y-label moved closer to the y-axis, 10px between is the recommendation for our use case or made configurable if different consumers have different requirements.

**Additional context**
Screenshoting showing ylabel far to the left of the y-axis:
image

AshishMotanamGurunadham:

-   [x] Feature
-   [ ] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

- Reduced gap in between y-axis and y-axis label when y2-axis is set to false.
 
**Dev site path example**
- https://engineering.cerner.com/carbon-graphs/#/line/simple

**Add any screenshots**

**Before**
image

**After**
image


src/test/unit/controls/Graph/GraphGenerate-spec.js
```diff
@@ -507,10 +507,39 @@ describe("Graph - Generate", () => {
                 axisDefault.x.label
             );
         });
-        it("Creates label y axis when text is present", () => {
+        it("Creates label y axis when y2-axis is false and when text is present", () => {
```
Veera81, line 510: nit: the wording little confusing. What is text present? Is that y-axis label present?

nit: It can be updated like below
```js
describe("when y2-axis is false and y-axis label is present", () => {
     it("should create y-axis label and reduce the gap between y-axis and y-axis label", () => {
    });
});
``

File: src/test/unit/controls/Graph/GraphGenerate-spec.js [hunk 2]
```diff
@@ -507,10 +507,39 @@ describe("Graph - Generate", () => {
...
+            expect(yLabelElement.querySelector("tspan").textContent).toBe(
+                axisDefault.y.label
+            );
+            expect(yLabelElement.getAttribute("transform")).toContain(
+                "rotate(-90)"
+            );
+            expect(toNumber(translate[0], 10)).toBeCloserTo(20);
+            expect(toNumber(translate[1], 10)).toBeCloserTo(115);
+        });
+        it("Creates label y axis when y2-axis is true and text is present", () => {
```
Veera81, line 528: nit: It can be updated like below
```js
describe("when y2-axis is true and y-axis label is present", () => {
     it("should create y-axis label and not reduce the gap between y-axis and y-axis label", () => {
    });
});
``

File: src/test/unit/controls/Graph/GraphGenerate-spec.js [hunk 3]
```diff
@@ -507,10 +507,39 @@ describe("Graph - Generate", () => {
...
+            ).translate;
+            expect(yLabelElement.nodeName).toBe("g");
+            expect(yLabelElement.querySelector("tspan").textContent).toBe(
+                axisDefault.y.label
+            );
+            expect(yLabelElement.getAttribute("transform")).toContain(
+                "rotate(-90)"
+            );
+            expect(toNumber(translate[0], 10)).toBeCloserTo(20);
+            expect(toNumber(translate[1], 10)).toBeCloserTo(115);
```
Veera81, line 526: Is this the exact difference between this (115) and the below test(120)?

Veera81: Never Mind. The 120 value belongs to different test case

src/test/unit/controls/Graph/GraphGenerate-spec.js
AshishMotanamGurunadham: Well that sounds good. But in order to maintain the consistency with previous test cases i just wrote it this way.

src/test/unit/controls/Graph/GraphGenerate-spec.js
AshishMotanamGurunadham: Well that sounds good. But in order to maintain the consistency with previous test cases i just wrote it this way.

PR #295, opened 2020-09-09
Fix: Region doesn't show face up (fixes #284 )

### LINKED ISSUE #284: Region doesn't show face up, when it has a content without values
**Describe the bug**
The region in the carbon-graphs does not show face up, even though there is only one line on the graph, if there is a content without values.

**Link to CodeSandbox**
Use [this](https://codesandbox.io/s/reverent-mendel-nphfx) template to reproduce the behavior.

**Steps to reproduce**
1. Load graph with a line graph with data containing region.
   * The line graph displays with region
2. Loan another line in the same graph/canvas which has empty array as values.
   * The region gets disappeared.

**Expected behavior**
The respective, region should appear where there is only one content in the graph.

**Screenshots**

-   OS: [e.g. iOS]
-   Browser [e.g. chrome, safari]

**Additional context**
The legend without data will be either disabled/hidden. So user will not get a chance to unselect the legend. So the graph will think, it has more than one content and hides all the regions.

SharathAlur:
Fixes: https://github.com/cerner/carbon-graphs/issues/284

-   [ ] Feature
-   [x] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

Avoid adding the graph to the `shownTargets` if the content do not have data.
Set `shouldHideAllRegion` to false if there is only one graph content displayed.

**Add any screenshots**
Screenshot 2020-09-09 at 3 19 17 PM
Screenshot 2020-09-09 at 3 20 25 PM


SharathAlur:
Sorry. I found there is an issue after rebasing with main branch. Once I fix that, I will ask for the review.

PR #301, opened 2020-09-16
Remove symbols with data set is removed from legend(Fixes #286).

### LINKED ISSUE #286: Remove symbols with data set is removed from legend.
Symbols are not being removed from the y-axis label when they are being removed from the legend due to reflow. If a data set has no data in the presented timeframe and is removed from the legend, the symbol should also be removed from the y-axis label. This will help eliminate some confusion for the user. 

What is happening: 
Screen Shot 2020-08-03 at 7 32 53 AM

What should happen: 
Screen Shot 2020-08-03 at 7 33 12 AM

AshishMotanamGurunadham:
Fixes: #286 

-   [ ] Feature
-   [x] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

- Added method to remove shape from y-axis when there is no values in data content during reflow in graphs like line, bar, paired result and scatter.
- Adjusted paired result data(for panning/paired result/dynamic data) to display content on the graph.

**Add any screenshots**

**Shapes in Line graph axis during reflow when there is no values in data content passed**
https://engineering.cerner.com/carbon-graphs/#/panning/line/y2-axis

**Before**
image

**After**
image

**Shapes in Scatter graph axis during reflow when there is no values in data content passed**
https://engineering.cerner.com/carbon-graphs/#/panning/scatter/dynamic-data

**Before**
image

**After**
image

**Shapes in Bar graph axis during reflow when there is no values in data content passed**
https://engineering.cerner.com/carbon-graphs/#/panning/bar/dynamic-data

**Before**
image

**After**
image

**Shapes in Paired result graph axis during reflow when there is no values in data content is passed**
https://engineering.cerner.com/carbon-graphs/#/panning/paired-result/dynamic-data

**Before**
image

**After**
image

**Paired result graph data display**
https://engineering.cerner.com/carbon-graphs/#/panning/paired-result/dynamic-data

**Before**
image

**After**
image


AshishMotanamGurunadham:
@jeremyfuksa I added all the data changes in this branch to see the desired outputs, You can just pull this branch, open the api's provided and click left or right chevron present in the page to see the changes.

jeremyfuksa:
On all dynamic data examples, panning brings up the "No Data" state, but that state does not go away once I pan back to where the previous data should be. Is that the intended behavior?

AshishMotanamGurunadham:
> On all dynamic data examples, panning brings up the "No Data" state, but that state does not go away once I pan back to where the previous data should be. Is that the intended behavior?

@jeremyfuksa Yes that is intended behavior.

jeremyfuksa:
> @jeremyfuksa Yes that is intended behavior.
OK. I'll approve.

jeremyfuksa (approved):

dev/examples/data.js
```diff
@@ -554,7 +554,7 @@ const DATA = [
                     upperLimit: 30
                 },
                 y2: {
-                    show: false,
+                    show: true,
```
mraplumb, line 557: Is there any negative consequence to other examples using this data object by changing these to `true`?

src/main/js/controls/Graph/Graph.js
```diff
@@ -418,7 +418,10 @@ class Graph extends Construct {
                 if (key === graphData.key) position = index;
             });
             if (position >= 0) {
-                if (this.content[position].type === "Bar") {
+                if (
+                    this.content[position].type === "Bar" &&
+                    graphData.values.length > 0
```
mraplumb, line 423: what is the value of this change?  are we trying to prevent `this.config.axis.x.ticks.values = [];` from executing when the length of the `graphData.values.length` is zero?

sdadn (approved):

Veera81:
> On all dynamic data examples, panning brings up the "No Data" state, but that state does not go away once I pan back to where the previous data should be. Is that the intended behavior?

I guess if we click on the previous arrow button, it should go back to the previous data

src/main/js/controls/Graph/Graph.js
AshishMotanamGurunadham: Yes Andy you are if we do not do so, we are getting an error saying that tick values are empty.

dev/examples/data.js
AshishMotanamGurunadham: Actually i added these changes to amke Jeremy work easy. Since he approved this PR i can goon and revert these chanegs.

Veera81 (approved):

PR #309, opened 2020-10-06
Fix: Fixed element.remove() issue with Internet Explorer(fixes #307)

### LINKED ISSUE #307: element.remove() function is not working in Internet explorer
**What part of Carbon does this bug correlate to?**
General graph logic:

* [x]  Graph attributes (axes, legends, regions...etc.)
* [ ]  Other (please specify): 

**Describe the bug**
- When we try to destroy one of the tooltip div element in Carbon-graphs, we are getting error like "**Object does not support property or method called remove**" because IE does not support **remove()** function for DOM elements.

**Expected behavior**
- div element should be removed during destroy without any errors in all browsers.

**Screenshots**
image

-   Browser [internet explorer]

AshishMotanamGurunadham:
Fixes : #307 

-   [ ] Feature
-   [x] Bug fix
-   [ ] Dependency updates
-   [ ] Documentation updates
-   [ ] Build related changes

-   [ ] Yes
-   [x] No

- Made changes to tooltip element remove method, so that it will also work in Internet explorer without any errors.


Veera81 (approved):

mraplumb:
@AshishMotanamGurunadham are there any screenshots or evidence that you can provide of this working in IE and the other supported browsers since this was an issue with browser compatibility?

src/main/js/helpers/label.js
```diff
@@ -338,7 +338,8 @@ const createTooltipDiv = () => {
  */
 const destroyTooltipDiv = () => {
     if (document.querySelector(`.${styles.labelPopupTooltip}`)) {
-        document.querySelector(`.${styles.labelPopupTooltip}`).remove();
+        const element = document.querySelector(`.${styles.labelPopupTooltip}`);
+        element.parentNode && element.parentNode.removeChild(element);
```
mraplumb, line 342: are we wanting to return `true` or possibly `false` from this?  If so, for what reason?

would it be more appropriate to use
```javascript
if(element.parentNode) {
   element.parentNode.removeChild(element);
}
```

AshishMotanamGurunadham:
> @AshishMotanamGurunadham are there any screenshots or evidence that you can provide of this working in IE and the other supported browsers since this was an issue with browser compatibility?

**Outputs when graph is destroyed:** 

**Internet Explorer:**
image

**Google chrome:**
image

**Firefox:** 
image

**Safari:**
image

src/main/js/helpers/label.js
AshishMotanamGurunadham: It won't return any boolean value Andy. It will just remove the element present in DOM. 

**element.parentNode && element.parentNode.removeChild(element);**
It is like kind of if statement if node is present then it will execute the logic else No and it won't make any difference either way i guess Andy.

mraplumb: I think if we are attempting to validate that `element.parentNode` exists and then retrieve the child node, I would prefer to do that with an explicit if block

Veera81 (approved):

src/main/js/helpers/label.js
AshishMotanamGurunadham: Fixed [here](https://github.com/cerner/carbon-graphs/pull/309/files#diff-ee13bfd863a6dec0adb29b9fc41caf73R342).

mraplumb (approved):

